PR #1, opened 2024-02-22
Workflow guidelines version 1

SimonHeybrock:
Add initial version of workflow guidelines.


docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+- Avoids keeping large data alive in memory if output metadata extraction depends on auxiliary input data or input metadata.
+
+### S.3: Avoid dependencies of output metadata on large data
+
+**Reason**
+Adding dependencies on large data to the output metadata extraction may lead to large data being kept alive in memory.
+
+**Note**
+Most of this is avoid by following S.2.
+A bad exampleAvoid would be to write to total raw counts to the output metadata, as this would require keeping the large data alive in memory, unless it is ensured that the task runs early.
```
jl-wynen, line 70:
```suggestion
A bad example would be writing the total raw counts to the output metadata, as this would require keeping the large data alive in memory, unless it is ensured that the task runs early.
```

jl-wynen, line 70: But how do we do this instead? This note just forbids writing certain things as metadata. But that cannot be avoided, e.g., in reflectometry as we need to write wavelength and theta ranges.

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+## S: Structure
+
+### S.1: Workflows shall be able to return final results as event data
+
+**Reason**
+- Required for polarization analysis, which wraps a base workflow.
+- Required for subsequent filtering, unless part of the workflow.
+
+**Note**
+There should be a workflow parameters (flag) to select whether to return event data or not.
```
jl-wynen, line 54:
```suggestion
There should be a workflow parameter (flag) to select whether to return event data or not.
```

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+This affects coordinates and data values independently.
+- If input counts are single-precision, the reduced intensity should be single-precision, and equivalently for double-precision.
+- If input coordinates are single-precision, derived coordinates should be single-precision, and equivalently for double-precision.
+
+**Note**
+This will allow for changing the precision of the entire workflow by choosing a precision when loading the input data.
+
+**Example**
+- If time-of-flight is single-precision, wavelength and momentum transfer should be single-precision.
+- In counts are single-precision, reduced intensity should be single-precision.
```
jl-wynen, line 88:
```suggestion
- If counts are single-precision, reduced intensity should be single-precision.
```

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+### S.3: Avoid dependencies of output metadata on large data
+
+**Reason**
+Adding dependencies on large data to the output metadata extraction may lead to large data being kept alive in memory.
+
+**Note**
+Most of this is avoid by following S.2.
+A bad exampleAvoid would be to write to total raw counts to the output metadata, as this would require keeping the large data alive in memory, unless it is ensured that the task runs early.
+
+### S.4: Handle single- and double-precision transparently
```
jl-wynen, line 72: I don't think 'transparent' is the right word here. How about 'Maintain floating point precision'?

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+### S.5: Switches to double-precision shall be deliberate, explicit, and documented
+
+**Reason**
+Some workflows may require switching to double-precision at a certain point in the workflow.
+This should be a deliberate choice, and the reason for the switch should be documented.
+
+### S.6: Propagation of uncertainties in broadcast operations should support "drop" and "upper-bound" strategies, "upper-bound" shall be the default
+
+**Reason**
+Unless explicitly computed, the exact propagation of uncertainties in broadcast operations is not well-defined.
```
jl-wynen, line 99: It is well defined but computationally intractable.

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+Dropping uncertainties is not desireable in general, as it may lead to underestimation of the uncertainties, but we realize that the upper-bound approach may not be suitable in all cases.
+We should therefore support two strategies, "drop" and "upper-bound", and "upper-bound" should be the default.
```
jl-wynen, line 101: Is this really something the user should toggle? How can the upper-bound strategy be valid for some data but not other for the same workflow?

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+See [Systematic underestimation of uncertainties by widespread neutron-scattering data-reduction software](http://dx.doi.org/10.3233/JNR-220049) for a discussion of the topic.
+TODO Add reference to upper-bound approach.
+
+### S.7: Do not write files or make write requests to services such as SciCat in providers
+
+**Reason**
+Providers should be side-effect free, and should not write files or make write requests to services.
+
+**Note**
+Workflow may run many times, or in parallel, or task may be retried after failure, and we want to avoid side-effects in these cases.
```
jl-wynen, line 113:
```suggestion
Workflows may run many times, or in parallel, or tasks may be retried after failure, and we want to avoid side-effects in these cases.
```

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+**Reason**
+Unless explicitly computed, the exact propagation of uncertainties in broadcast operations is not well-defined.
+Dropping uncertainties is not desireable in general, as it may lead to underestimation of the uncertainties, but we realize that the upper-bound approach may not be suitable in all cases.
+We should therefore support two strategies, "drop" and "upper-bound", and "upper-bound" should be the default.
+
+**Note**
+See [Systematic underestimation of uncertainties by widespread neutron-scattering data-reduction software](http://dx.doi.org/10.3233/JNR-220049) for a discussion of the topic.
+TODO Add reference to upper-bound approach.
+
+### S.7: Do not write files or make write requests to services such as SciCat in providers
```
jl-wynen, line 107: Not sure we should keep SciCat requests out of the pipeline. This adds additional code around the pipeline and removed the relevant parameters from the purview of the pipeline (it can only track local filenames which are not unique instead of unique SciCat IDs)

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+## T: Testing
+
+### T.1: Adherence to the guidelines shall be tested, and the guideline ID shall be referenced in the test name
+
+**Reason**
+We want to ensure that the guidelines are followed, and that this remains the case as the code base evolves.
+Referencing the guideline ID in the test name makes it easier to find the relevant guideline (or vice-versa), or remove the test if the guideline is removed.
+
+**Note**
+Note all guidelines are testable.
```
jl-wynen, line 133:
```suggestion
Not all guidelines are testable.
```

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
...
+
+### T.2: Write unit tests for providers
+
+**Reason** Unit tests for providers are easier to write and maintain than for entire workflows.
+
+**Note** This does not mean that we should not write tests for entire workflows, but that we should also write tests for providers.
```
jl-wynen, line 139: How about a guideline for structuring the test directory? And for how to select whether to run integration tests or not. If they take a long time to run, it would be useful to be able to easily deselect them.

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: I am pretty sure that at least the IDS or IS will want to toggle this.

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: I was specifically referring to write-requests, would you want to much such requests?

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: I think splitting unit tests from integration tests (or something like that) is a great suggestion! As we have not done that before, maybe we can add it in a future version, once we have tried it out in a project?

docs/user-guide/reduction-workflow-guidelines.md
jl-wynen: Somehow I did not register the 'write'. Never mind then

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -0,0 +1,139 @@
+# Reduction Workflow Guidelines
```
nvaytet, line 1: Nothing to add from my side :+1:

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: There will be exceptions, not everything can be "avoided", this is why I chose this formulation instead of making it stricter.

SimonHeybrock:
@jl-wynen @nvaytet I think I addressed all comments. I have cleaned this up and suggest releasing a first agreed-upon version such that we can begin implementation. I included a listing of the remaining items which we can add subsequently.

docs/user-guide/reduction-workflow-guidelines.md
jl-wynen: Then I'd like clarification about how strict these guidelines are. As it is written, it sounds like a command. So if this one may be violated, can all rules be broken if we can argue for it?

SimonHeybrock: Time will tell? I think we should be agile and avoid over-analyzing for now, see what works in practice and what does not, then iterate and improve.

SimonHeybrock: Also, in the case you mentioned, shouldn't the wavelength and theta ranges report the *used* ranges, instead of those in the raw data? I presume the workflow may limit this, and in any case it can be computed from derived data instead of raw data, limiting the problem.

jl-wynen: > Also, in the case you mentioned, shouldn't the wavelength and theta ranges report the _used_ ranges, instead of those in the raw data? I presume the workflow may limit this, and in any case it can be computed from derived data instead of raw data, limiting the problem.

Possibly. The current workflow does not allow limiting theta. Also, what if the bounds are much larger than the data? I'd want to report the limits of the data that is actually used instead.

PR #6, opened 2024-02-26
Clarify monitor and detector handling guidelines

SimonHeybrock:
In this case I updated some guidelines IDs, which we should generally avoid. As this was just merged this morning and is not implemented anywhere I think this should be ok.

### DIFF
--- a/docs/user-guide/reduction-workflow-guidelines.md
+++ b/docs/user-guide/reduction-workflow-guidelines.md
@@ -78,23 +78,36 @@ This is often not apparent from small test data, as the location of performance 
 **Note**
 There should be a workflow parameter (flag) to select whether to return event data or not.
 
-### S.2: Split loading and handling of monitors and detectors from loading of auxiliary data and metadata
+### S.2: Load each required NXmonitor separately
 
 **Reason**
-- Allows for more efficient parallelism and reduction in memory use.
-- Avoids loading large data into memory that is not needed for the reduction.
-- Avoids keeping large data alive in memory if output metadata extraction depends on auxiliary input data or input metadata.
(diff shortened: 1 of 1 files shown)


docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -78,23 +78,36 @@ This is often not apparent from small test data, as the location of performance
...
 
-### S.2: Split loading and handling of monitors and detectors from loading of auxiliary data and metadata
+### S.2: Load each required NXmonitor separately
 
 **Reason**
-- Allows for more efficient parallelism and reduction in memory use.
-- Avoids loading large data into memory that is not needed for the reduction.
-- Avoids keeping large data alive in memory if output metadata extraction depends on auxiliary input data or input metadata.
+Monitor data can be extremely large when operating in event mode.
+Loading only individual monitors avoids loading unnecessary data and allows for more efficient parallelism and reduction in memory use.
```
jl-wynen, line 85: Not really related to guidelines, but what about transforms? Are there cases where we need to load more than just a monitor in order to find the position? Same for detectors.

SimonHeybrock: This can be handled by the loader. We should probably put this into ESSreduce (or even ScippNexus), and just wrap that elsewhere.

jl-wynen: That means reimplementing the logic to walk depends-on chains. So it seems to be something that should be handled by `compute_positions` or a similar function in ScippNexus.

jl-wynen (approved):

PR #9, opened 2024-03-04
Guideline rules on name and type conventions

### LINKED ISSUE #7: Naming conventions
Collecting a list of conventions. Feel free to edit the post.

#### Names used in ESSsans that can be used by other packages:
- Filename, DirectBeamFilename (simple filename to be processed by, e.g., Pooch, Scitacean)
- FilePath (path to an actual file on disk)
- SampleRun, BackgroundRun, EmptyBeamRun (run type ending in 'Run')
- RunType (TypeVar constrained to the above run types)
- RunTitle, RunNumber (helper IDs to be converted to/from RunType. Pipeline should use RunType to parametrise generics)
- MonitorType
- NeXus* (all keys referring to NeXus file contents use this spelling)
  - NeXus*Name [NeXusSampleName, NeXusSourceName, NeXusDetectorName] (names, *not* full paths, of groups in a NeXus file)
- SamplePosition, SourcePosition
- *Bins [WavelengthBins, DspacingBins, QBins] (variable of bin-edges)
- UncertaintyBroadcastMode (enum flag)
- ReturnEvents (bool flag)
- CorrectForGravity (bool flag)

#### Up for discussion:
- Incident, Transmission (used in ESSsans for monitors but ambiguous, could also refer to beams / intensities)

#### Inconsistencies:
- OutFilename (in sans: used as concrete path, so inconsistent with input Filename, FilePath)
- EmptyBeamRun vs (DirectBeam, DirectDeamFilename)

jl-wynen:
Fixes #7 

Please flag any and all names / types that you don't think are appropriate! And suggest any that I missed.


docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -43,6 +40,56 @@ We plan to include the following in future versions of the guidelines:
...
+| Filename \| *Filename         | str         | Simple name of a file, must be processed into FilePath                    |
+| FilePath \| *FilePath         | Path        | Concrete path to a file on the host filesystem, ideally absolute          |
+| --- **Run IDs** ---           |             |                                                                           |
+| SampleRun, BackgroundRun, ... | Any         | Identifier for a run                                                      |
+| RunType                       | TypeVar     | Constrained to the run types used by the package, see above               |
+| RunTitle                      | str         | Extracted from NeXus or provided by user, can be used to find files       |
+| --- **Monitors** ---          |             |                                                                           |
+| IncidentMonitor \| *Monitor   | Any         | Identifier for a monitor                                                  |
+| MonitorType                   | TypeVar     | Constrained to the monitor types used by the package, see above           |
```
SimonHeybrock, line 64: Here we assume that a workflow necessarily wants/needs to use generics for runs types and monitor types. Is that a good assumption? Or should we consider formulating it as "if using generics for X, use this convention"?

Should we maybe split `C.1` to discuss naming for generics (and their typevars) separately?

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -43,6 +40,56 @@ We plan to include the following in future versions of the guidelines:
...
+  ```python
+  P = NewType('P', numbers.Real)
+  pipeline[P] = 3.0  # works
+  pipeline[P] = 3    # works, too, but not if `P` were `float`
+  ```
+- Use {class}`collections.abc.Sequence` instead of `list` or `tuple`.
+  - But do *not* use {class}`typing.Iterable`!
+    Parameters may be consumed multiple times and iterables are not guaranteed to support that.
+- Gracefully promote dtypes for small parameters.
+  E.g., `sc.scalar(2, unit='m')` and `sc.scalar(2.0, unit='m')` should be usable interchangeably.
```
SimonHeybrock, line 91: 👍 
Should we require the same, e.g., for params like bin-edges such that, e.g., passing `arange` works just like passing `linspace`?

docs/user-guide/reduction-workflow-guidelines.md
jl-wynen: > Here we assume that a workflow necessarily wants/needs to use generics for runs types and monitor types. Is that a good assumption? Or should we consider formulating it as "if using generics for X, use this convention"?

I don't understand what you are saying. When we don't use generics, the SampleRun, ... and RunType simply don't exist. This table doesn't mandate using all entries, it's supposed to tell us how to name them if and when we need them.

> Should we maybe split C.1 to discuss naming for generics (and their typevars) separately?

Why? How would you structure that table?

SimonHeybrock: > I don't understand what you are saying. When we don't use generics, the SampleRun, ... and RunType simply don't exist.

Say I want to write a workflow that process a sample run and a vanadium run, without use of generics. Which names do I use?

I am suggesting to make two guidelines (each with their own table), or have a separate guideline specifying conventions for naming typevars and identifiers.

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: Operations may raise when ints are passed instead of floats as coords (bin edges, for example). I am suggesting extending your suggestion for graceful handling.

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: If the reader is not familiar with using generic providers in a workflow then the current table is quite unclear. Please make a separate guidelines. We do not need to make an explicit list.

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -43,6 +40,80 @@ We plan to include the following in future versions of the guidelines:
...
+If a workflow uses generics to parametrize its types, e.g., `Filename`,
+it should define new types used as tags and type vars constrained to those tags.
+
+**Table**
+
+| Name                                      | Type    | Description                                                     |
+|-------------------------------------------|---------|-----------------------------------------------------------------|
+| --- **Run IDs** ---                       |         |                                                                 |
+| SampleRun, BackgroundRun, ...             | Any     | Identifier for a run, only used as a type tag                   |
+| RunType                                   | TypeVar | Constrained to the run types used by the package, see above     |
```
SimonHeybrock, line 82: So is the rule to end typevars with `Type`?

jl-wynen: Don't know in general. I've only seen those two in practice.

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -43,6 +40,80 @@ We plan to include the following in future versions of the guidelines:
...
+**Reason**
+Helps with sticking to established practices and working across packages.
+
+**Table**
+Names use glob syntax, i.e., '*Filename' is any string that ends in 'Filename'.
+
+| Name                        | Type        | Description                                                               |
+|-----------------------------|-------------|---------------------------------------------------------------------------|
+| --- **Files** ---           |             |                                                                           |
+| Filename \| *Filename       | str         | Simple name of a file, must be processed into FilePath                    |
```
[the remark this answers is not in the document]
jl-wynen: SciCat datasets can contain any number of files. So we need a pair or (dataset id, file id). And the file id can only really be a file name.

docs/user-guide/reduction-workflow-guidelines.md
jl-wynen: @SimonHeybrock Do we need to do anything about this?

SimonHeybrock: Should we make this a rule? I think it is quite confusing to know what is a typevar and what is and ID otherwise?

SimonHeybrock (approved):

PR #10, opened 2024-03-06
Generic NeXus loaders

### LINKED ISSUE #8: NeXus load helpers
Add a `ess.reduce.nexus` sumodule with the wollowing:

- `nexus.Filename` (not parametrized)
- `nexus.load_monitor`, `nexus.MonitorName` (not parametrized!), loading `nexus.RawMonitor` from a single NXmonitor
- `nexus.load_detector`, `nexus.DetectorName` (not parametrized!), loading `nexus.RawDetector` from a single NXdetector
- ~`nexus.load_auxiliary`, everything by not monitors and detectors. Source and sample position can be extracted from that.~
  - Subtlety: If we load all this, we cannot identify source and sample anymore. Solution?
    - ~Identify source and sample position already here?~
    - ~Or make ScippNexus include the `NX_class` attribute as a datagroup item, so we can do this after loading.~
    - `load_auxiliary` is a bad idea, and we should have `load_source` and `load_sample` instead!
    - Address other metadata in follow-up tasks, once this has been shown to be functional, e.g., by using it in ESSsans.
- Use ScippNexus' feature of finding child classes by type, instead of relying on names (see https://scipp.github.io/scippnexus/user-guide/quick-start-guide.html#Getting-all-children-of-a-specific-NX_class). Example suggestion:
  ```python
  def _unique_child(group: snx.Group, child: Type[snx.NXobject]) -> snx.Group:
      children = group[child]
      if len(children) != 1:
          raise ValueError(f'Expected exactly one {child} group.')
      return next(iter(children.values()))
  ```
- Note (and fix) scipp/esssans#109 and scipp/esssans#110.
- Should all the above optionally work with an open file? This would also required for seamless operation with Beamlime, where we have a JSONGroup (defined in ScippNeutron) instead of a file(name).
- Extra parameters for, e.g., selecting time ranges will be added later. Do not implement such features in this initial iteration.

jl-wynen:
Fixes #8

They are heavily based on the LOKI loaders in ESSsans and the test files used in that package.


src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,329 @@
...
+        return child
+
+    children = group[nx_class]
+    if len(children) != 1:
+        raise ValueError(f'Expected exactly one {nx_class} group, got {len(children)}')
+    return next(iter(children.values()))  # type: ignore[return-value]
+
+
+def extract_detector_data(
+    detector: RawDetector, detector_name: DetectorName
```
SimonHeybrock, line 267:
```suggestion
    detector: RawDetector
```
I don't think we should have `DetectorName` here. Instead return the event data, if found, otherwise the child data array (dense data). Reason: I don't think we should rely on the detector name being related to the name of the event-data subgroup. In the case of monitors there may be no subgroup at all. Also it is simpler to just look for the child data array(s).

src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,329 @@
...
+    --------
+    load_detector:
+        Load a detector from a NeXus file in a format compatible with
+        ``extract_detector_data``.
+    """
+    return RawDetectorData(_extract_events_or_histogram(detector, detector_name))
+
+
+def extract_monitor_data(
+    monitor: RawMonitor, monitor_name: MonitorName
```
SimonHeybrock, line 293:
```suggestion
    monitor: RawMonitor
```
dito

src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,329 @@
...
+    --------
+    load_monitor:
+        Load a monitor from a NeXus file in a format compatible with
+        ``extract_monitor_data``.
+    """
+    return RawMonitorData(_extract_events_or_histogram(monitor, monitor_name))
+
+
+def _extract_events_or_histogram(dg: sc.DataGroup, name: str) -> sc.DataArray:
+    data_names = {f'{name}_events', 'data'}
```
SimonHeybrock, line 319: I suggest to remove the reliance on naming. Just look for child data arrays, prefer to return the event-data one?

src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,329 @@
...
+"""Raw data from a NeXus sample."""
+RawSource = NewType('RawSource', sc.DataGroup)
+"""Raw data from a NeXus source."""
+
+
+def load_detector(
+    file_path: Union[FilePath, NeXusFile, NeXusGroup],
+    *,
+    detector_name: DetectorName,
+    instrument_name: Optional[InstrumentName] = None,
```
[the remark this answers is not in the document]
jl-wynen: Sounds reasonable

src/ess/reduce/nexus.py
jl-wynen: How can I identify the data? It's just one of many data arrays in the group.

jl-wynen: Or can there only be one data array per detector / monitor?

src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,345 @@
...
+    nx_class: Type[snx.NXobject],
+    entry_name: Optional[EntryName] = None,
+) -> sc.DataGroup:
+    with _open_nexus_file(file_path) as f:
+        entry = _unique_child_group(f, snx.NXentry, entry_name)
+        instrument = _unique_child_group(entry, snx.NXinstrument, None)
+        loaded = cast(
+            sc.DataGroup, _unique_child_group(instrument, nx_class, group_name)[()]
+        )
+        loaded = snx.compute_positions(loaded)
```
SimonHeybrock, line 237: We have to support the `store_transform` arg. The transforms are required in SANS (for the detectors). Probably for now we can just always enable this, we can revisit later if it turns out there are issues with it.

src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,345 @@
...
+            "Raw data loaded from NeXus does not contain events or a histogram. "
+            "Expected to find a data array, "
+            f"but the data only contains {set(dg.keys())}"
+        )
+    if len(data_arrays) > 1:
+        raise ValueError(
+            "Raw data loaded from NeXus contains more than one data array. "
+            "Cannot uniquely identify the event or histogram data. "
+            f"Got items {set(dg.keys())}"
+        )
```
SimonHeybrock, line 344: This won't do, unfortunately. Facilities have the tendency of storing a histogrammed "preview" of the event data alongside the events. I think we should return the events if found, and the dense data otherwise.

src/ess/reduce/nexus.py
jl-wynen: They are stored in the data group. Does the sans code need to change the name of the item?

tests/nexus_test.py
```diff
@@ -0,0 +1,306 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
nvaytet, line 1: Should we try (maybe as a pair-programming thing?) to test this on the loki workflow?

src/ess/reduce/nexus.py
SimonHeybrock: We already agreed to just store it as "transform".

jl-wynen:
> We need to know where to find it, so we relied on a name that was defined in the default parameters.
>
> I guess it could either be a parameter passed to this, or maybe it's enough to just have always the same name hard-coded here, but we need to make sure we never have a name clash in a file?

Added a check as for transformation. When discussing with Simon, we concluded to just user hard coded names for now because the chance of a conflict is very low. We can always add an argument later if need be.

src/ess/reduce/nexus.py
```diff
@@ -0,0 +1,345 @@
...
+RawMonitorData = NewType('RawMonitorData', sc.DataArray)
+"""Data extracted from a RawMonitor."""
+RawSample = NewType('RawSample', sc.DataGroup)
+"""Raw data from a NeXus sample."""
+RawSource = NewType('RawSource', sc.DataGroup)
+"""Raw data from a NeXus source."""
+
+
+def load_detector(
+    file_path: Union[FilePath, NeXusFile, NeXusGroup],
```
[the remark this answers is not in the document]
jl-wynen: It works as long as any type is provided. So, this can be used in a pipeline as is. But I think all production pipelines will need different kinds of files. So we will have to wrap it.

tests/nexus_test.py
jl-wynen: Good idea!

nvaytet: We tested and it works nicely: see https://github.com/scipp/esssans/pull/114

src/ess/reduce/nexus.py
```diff
@@ -234,7 +239,15 @@ def _load_group_with_positions(
         loaded = cast(
             sc.DataGroup, _unique_child_group(instrument, nx_class, group_name)[()]
         )
-        loaded = snx.compute_positions(loaded)
+
+        transform_out_name = 'transformation'
```
SimonHeybrock, line 243: I think we were using `transform` in ESSsans? `transformation` is nearly identical to `transformations`, the current name used for the subgroup holding the transformations, so this would be quite confusing. Can we stick with `transform`?

SimonHeybrock (approved):

PR #18, opened 2024-04-08
feat: add provider for downloading datasets from scicat

### LINKED ISSUE #15: Mixing local files and files from SciCat
@wpotrzebowski pointed out that we currently cannot mix files downloaded with pooch with local files. It is all or none.

While we do not plan to support this for pooch (since it is only for testing and documentation), this is an important requirement when working with SciCat. For example, mask files or other references may be obtained from SciCat, while a user wants to process a local file.

jokasimr:
Fixes #15


src/ess/reduce/scicat.py
```diff
@@ -0,0 +1,31 @@
...
+
+from .nexus import FilePath
+
+
+def download_scicat_file(
+    dataset_id: str,
+    filename: str,
+    *,
+    token: str,
+    version: Optional[str] = None,
```
jl-wynen, line 15: Better parametrize over the client. This allows users to construct their own if they have special needs (e.g., special SSH auth) and we can use a fake in tests.

tests/scicat_test.py
```diff
@@ -0,0 +1,28 @@
...
+from ess.reduce.scicat import download_scicat_file
+
+
+class Client:
+    @staticmethod
+    def from_token(url, token, file_transfer):
+        return setup_fake_client()
+
+
+ess.reduce.scicat.Client = Client
```
jl-wynen, line 16: See my comment about parametrizing the download function.

tests/scicat_test.py
```diff
@@ -0,0 +1,28 @@
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from scitacean.testing.docs import setup_fake_client
+
+import ess
+from ess.reduce.scicat import download_scicat_file
+
+
+class Client:
```
jl-wynen, line 10: Use https://scicatproject.github.io/scitacean/generated/modules/scitacean.testing.backend.fixtures.fake_client.html instead. (This is a pytest fixture)

src/ess/reduce/scicat.py
jl-wynen, line 1: How do you want to use this function? It's not a provider and quite restrictive in how it wants to download data. In a real workflow, we need the dataset (metadata) along with the file. And we may also need more than one file per dataset.
So I would split it and have a provider that downloads a dataset and a separate one that downloads files.

jl-wynen: It's definitely need to be wrapped, just like the nexus loaders. This means that the basic ones here will be very simple. So I think the main task here is figuring out what parameters should be requested.

pyproject.toml
```diff
@@ -32,6 +32,7 @@ requires-python = ">=3.10"
 dependencies = [
     "scipp >= 24.02.0",
     "scippnexus >= 24.03.0",
+    "scitacean[sftp, test]",
```
SimonHeybrock, line 35: If we avoid using extras then the new way of defining `meta.yaml` automatically (see `copier_template`) will work. Also, why do we depend on the `test` extra, that seems odd?

src/ess/reduce/scicat.py
```diff
@@ -0,0 +1,48 @@
...
+    dataset_id: str,
+    filename: str,
+    *,
+    client: Client,
+    target: Optional[Path] = None,
+) -> FilePath:
+    if target is None:
+        target = Path(f'~/.cache/essreduce/{dataset_id}')
+    dset = client.get_dataset(dataset_id)
+    dset = client.download_files(dset, target=target, select=filename)
```
SimonHeybrock, line 45: Strictly speaking this is a "side effect", which we recommend to avoid... but if I understand correctly then this is an exception because it is more like caching a file?

src/ess/reduce/scicat.py
```diff
@@ -0,0 +1,48 @@
...
+    if version is None:
+        version = 'v3'
+    return Client.from_token(
+        url=f"https://scicat.ess.eu/api/{version}",
+        token=token,
+        file_transfer=file_transfer,
+    )
+
+
+def download_scicat_file(
```
SimonHeybrock, line 35: - Will this re-download a file if it has been fetched before?
- Is this provider thread-safe? What if Dask calls it multiple times?

pyproject.toml
jokasimr: Yes I'll remove the `test` extra. How can we avoid depending on the `sftp` extra? By manually adding the dependencies?

pyproject.toml
```diff
@@ -32,7 +32,7 @@ requires-python = ">=3.10"
 dependencies = [
     "scipp >= 24.02.0",
     "scippnexus >= 24.03.0",
-    "scitacean[sftp, test]",
+    "scitacean[sftp]",
```
SimonHeybrock, line 35: Can we avoid using extras?

src/ess/reduce/scicat.py
```diff
@@ -41,6 +44,7 @@ def download_scicat_file(
 ) -> FilePath:
     if target is None:
         target = Path(f'~/.cache/essreduce/{dataset_id}')
-    dset = client.get_dataset(dataset_id)
-    dset = client.download_files(dset, target=target, select=filename)
+    with _scicat_download_lock:
+        dset = client.get_dataset(dataset_id)
+        dset = client.download_files(dset, target=target, select=filename)
```
SimonHeybrock: I suppose this will break when multiple process / nodes are used that share the same filesystem?

jokasimr: Yes it will not work in that situation.
If the user pre-downloads the datasets they need then `client.download_files` will not download anything (right? @jl-wynen) and then there won't be any race condition. Is that good enough?

What kind of solution would you like to see @SimonHeybrock ?

SimonHeybrock: I do not have any specific solution in mind. I just think that we should take a step back and see if we can take a different approach (such as running something *before* the pipeline) that can avoid the problem, instead of trying to fix the symptoms.

jokasimr: Downloading the data before running the pipeline is already an option. The user can do that using scitacean or calling the function here directly.

I'll leave this PR as is for now until we figure out what kind of solution we want here.

jl-wynen:
We need to be careful with how we pass datasets and files to pipelines, see the service blueprint (https://project.esss.dk/nextcloud/index.php/apps/files/?dir=%2F&openfile=18026378). This will require prefetching files and providing their path or a dataset that contains the local path as parameters.

Based on this, I don't see much utility in the simple functions implemented in this PR. @jokasimr can you check how they fit into the blueprint? If they don't I would suggest closing this PR.

jokasimr:
In the blueprint I think the parts that are relevant to this pr are the following entries:
1. request a datasets by id. 
2. extracting relevant related datasets.
3. request files that are not already local or out of date.

(1) already exists in scitacean and I don't think it makes sense to wrap that further.
(2) and (3) are implemented in this PR.

This is a bare-bones implementation, sure, but I don't think it makes sense to build much more before we've tried integrating scicat datasets/files in the workflows and discovered what more is needed.

Maybe it's easier to talk about this in person so you can explain more what kind of solution you had in mind for this issue.

SimonHeybrock:
Please check out scipp/esssans#136 and in particular the addition of helper functions in https://github.com/scipp/esssans/pull/136/files#diff-21119170cad13c69e1a8e14143c2527950f4aa5308d96851c51a3fc280faff1d such as `get_sans2d_tutorial_data_folder`, which runs *before* the workflow and handles downloading files. Can we take a similar approach with Scicat data?

jl-wynen:
`get_sans2d_tutorial_data_folder` is specialised to the tutorial. We need something more general to work with user data. But yes, we may be able to write a function that takes a dataset ID and filter to identify the data file and downloads the dataset, file, and all required associated files.

SimonHeybrock:
> `get_sans2d_tutorial_data_folder` is specialised to the tutorial. We need something more general to work with user data. But yes, we may be able to write a function that takes a dataset ID and filter to identify the data file and downloads the dataset, file, and all required associated files.

I think my question really was: Would such an approach fit into the user story / service blueprint?

jl-wynen:
> I think my question really was: Would such an approach fit into the user story / service blueprint?

Essentially, yes. This is what the user story looks like. 1. Select and download the datasets + files. 2. provide them to the workflow.

PR #22, opened 2024-04-24
Add guideline on conventional pattern for code-based user interface

### LINKED ISSUE #21: UX improvements for our workflows
### Context

In a recent call with an instrument scientist it became clear that our notebooks of reduction workflows are unnecessarily unclear for non-programmers, even when considering an expert user who *does* want to see and access all the details.

Based on my notes, here are a couple of suggestions of potential improvements.

### Suggestions

1. Replace "providers" with "workflow steps", avoiding Sciline terminology.
2. Replace "pipeline" with "reduction".
3. Avoid param dicts: Mixing of collecting/setting params in a dict as well as setting them on a `Pipeline` is confusing. It is not obvious to users that this is equivalent.
4. Add wrappers for pipeline creation that include defaults, mainly to facilitate 3.) above.

### Example

Change
```python
params = loki.default_parameters()
params[QBins] = ...
pipeline = sciline.Pipeline(loki.providers, params=params)
pipeline[ReturnEvents] = True
```
to
```python
reduction = loki.LokiReduction()
reduction[QBins] = ...
reduction[ReturnEvents] = True
```
If customization of providers is required, we can have something like
```python
reduction = sans.Reduction(
    steps=loki.workflow_steps + sans.workflow_steps,
    params=loki.default_parameters())
```

SimonHeybrock:
Fixes #21.

### DIFF
--- a/docs/user-guide/reduction-workflow-guidelines.md
+++ b/docs/user-guide/reduction-workflow-guidelines.md
@@ -122,7 +122,7 @@ Users should not have to worry about the concrete type of parameters.
   E.g., `sc.scalar(2, unit='m')` and `sc.scalar(2.0, unit='m')` should be usable interchangeably.
   This can also apply to arrays, for instance, `sc.linspace` and `sc.arange` should be interchangeable but the latter may result in integers while the former typically produces floats.
 
-### C.5: Use a conventional pattern minimizing programming jargon, favor fewer and concepts
+### C.5: Use a fixed pattern for creating, manipulating, and running workflows
 
 **Reason**
 - Terms such as provider or pipeline mean nothing to users, using their language instead reduces cognitive load.
--- a/docs/user-guide/reduction-workflow-guidelines.md
+++ b/docs/user-guide/reduction-workflow-guidelines.md
@@ -124,7 +124,8 @@ Users should not have to worry about the concrete type of parameters.
 ### C.5: Use a fixed pattern for creating, manipulating, and running workflows
 
 **Reason**
-- Terms such as provider or pipeline mean nothing to users, using their language instead reduces cognitive load.
+- Using terms such as provider or pipeline increases cognitive load for scientific users.
+  The reason is that those terms are unfamiliar to users based on the scientific domain language they are used to.
 - Manipulating multiple concepts such as (1) a dict of parameters, (2) a `sciline.Pipeline`, and (3) a `sciline.TaskGraph` is confusing, especially for non-programmers.
 
 **Notes**
--- a/docs/user-guide/reduction-workflow-guidelines.md
+++ b/docs/user-guide/reduction-workflow-guidelines.md
@@ -122,6 +122,31 @@ Users should not have to worry about the concrete type of parameters.
   E.g., `sc.scalar(2, unit='m')` and `sc.scalar(2.0, unit='m')` should be usable interchangeably.
   This can also apply to arrays, for instance, `sc.linspace` and `sc.arange` should be interchangeable but the latter may result in integers while the former typically produces floats.
 
+### C.5: Use a conventional pattern minimizing programming jargon, favor fewer and concepts
+
+**Reason**
+- Terms such as provider or pipeline mean nothing to users, using their language instead reduces cognitive load.
+- Manipulating multiple concepts such as (1) a dict of parameters, (2) a `sciline.Pipeline`, and (3) a `sciline.TaskGraph` is confusing, especially for non-programmers.
+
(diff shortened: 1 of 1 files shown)


docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -122,6 +122,31 @@ Users should not have to worry about the concrete type of parameters.
...
+  - Prefix with `make` and suffix with `workflow`, qualifiers can be added in between, for example `make_abc_workflow`.
+- Avoid creating parameter dicts in notebooks, set parameters on the workflow object directly.
+- Avoid calling `workflow.get` (which would return a `sciline.TaskGraph`), instead call `workflow.compute` and `workflow.visualize`, even if it means listing the result and building the task graph multiple times.
+
+**Example**
+
+```python
+from ess import loki
+
+workflow = loki.make_workflow()
```
SimonHeybrock, line 143: I am tempted to change this to a camel-case function name (`loki.LokiWorkflow()` or sth. like that). Wouldn't it make it easier to switch to a class later, without having to touch all the notebook?

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -122,6 +122,31 @@ Users should not have to worry about the concrete type of parameters.
   E.g., `sc.scalar(2, unit='m')` and `sc.scalar(2.0, unit='m')` should be usable interchangeably.
   This can also apply to arrays, for instance, `sc.linspace` and `sc.arange` should be interchangeable but the latter may result in integers while the former typically produces floats.
 
+### C.5: Use a conventional pattern minimizing programming jargon, favor fewer and concepts
```
jl-wynen, line 125: Not sure how you mean this sentence. Is it
```suggestion
### C.5: Use a conventional pattern minimizing programming jargon, favor fewer concepts
```
or
```suggestion
### C.5: Use a conventional pattern minimizing programming jargon and concepts and favor fewer concepts
```
?

In any case, I would be more specific:
```suggestion
### C.5: Use a fixed pattern for creating, manipulating, and running workflows
```

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -122,6 +122,31 @@ Users should not have to worry about the concrete type of parameters.
...
+
+**Reason**
+- Terms such as provider or pipeline mean nothing to users, using their language instead reduces cognitive load.
```
jl-wynen, line 128:
```suggestion
- Terms such as provider or pipeline mean nothing to users, using their natural language and instead reduces cognitive load.
```

SimonHeybrock: Their natural language is English, I was referring to how they are used to referring to things in their work/research.

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: @jl-wynen Thoughts on this one?

SimonHeybrock:
See scipp/esssans#136 for how this looks in practice.

SimonHeybrock:
Updated based on experience in scipp/esssans#136. In particular I removed `FilePath` and use camel case for creating workflows.

docs/user-guide/reduction-workflow-guidelines.md
jl-wynen: You don't have to touch the notebooks. It would still be possible to have a function that instantiates the class. Then there would be two ways of doing the same thing which would not be great. But it wouldn't be a big issue either, IMHO.

SimonHeybrock: Ok, then let us leave it as it is.

docs/user-guide/reduction-workflow-guidelines.md
SimonHeybrock: That does not parse grammatically.

jl-wynen: True. Let's be more explicit:
```
Using terms such as provider or pipeline increases cognitive load for scientific users.
The reason is that those terms are unfamiliar to users based on the scientific domain language they are used to.
```

SimonHeybrock:
```suggestion
- Using terms such as provider or pipeline increases cognitive load for scientific users.
  The reason is that those terms are unfamiliar to users based on the scientific domain language they are used to.
```

nvaytet (approved):

PR #26, opened 2024-05-07
Migrate masking a range from sans

nvaytet:
This was created in `esssans` but I would like to make use of it in `essdiffraction` also.


nvaytet:
Not sure I need this anymore, since we are trying to use masks defined as functions in https://github.com/scipp/essdiffraction/pull/55

nvaytet:
Closing as this is stale and probably outdated. We can always re-open later if needed.

PR #30, opened 2024-05-23
Add interactive masking tool

nvaytet:
Part of https://github.com/scipp/essdiffraction/issues/50

- Masks can be drawn on 1d and 2d data (using `mpltoolbox`).
- They can then be saved to file, using Scipp's hdf5 format.
- Only rectangular and horizontal/vertical spans are supported for now.

Screenshot at 2024-05-24 00-06-34
Screenshot at 2024-05-24 00-05-51


nvaytet:
To be moved to `scippneutron`

PR #36, opened 2024-06-12
Add nexus definition option to loaders 

### LINKED ISSUE #33: Add possibility to pass definitions to nexus loaders
In `essdiffraction`, when loading dream files, we override some definitions to avoid loading the Off geometry (see [here](https://github.com/scipp/essdiffraction/blob/a9941fb63b29ac8d7f8e419d9126d557bca6dc05/src/ess/dream/io/nexus.py#L61))
When moving to using the `essreduce` nexus utilities, the ability to do this is gone.
We should allow passing definitions in our utilities here.

jokasimr:
Fixes #33 

There are a couple of issues remaining:

* If the `file_path` argument is an already opened nexus file,
  * can we add the definitions to that object (requires changes in scippnexus?)
  * or do we raise an error?
* What kind of definitions should we use in the tests?

### DIFF
--- a/src/ess/reduce/nexus.py
+++ b/src/ess/reduce/nexus.py
@@ -293,7 +293,10 @@ def _open_nexus_file(
     definitions: Optional[Mapping] = None,
 ) -> ContextManager:
     if isinstance(file_path, getattr(NeXusGroup, '__supertype__', type(None))):
-        # Bug. Definitions ignored if the file_path is an already opened nexus file.
+        if definitions is not None:
+            raise ValueError(
+                "Cannot apply new definitions to open nexus file or nexus group."
+            )
         return nullcontext(file_path)
     if definitions is not None:
         return snx.File(file_path, definitions=definitions)
--- a/src/ess/reduce/nexus.py
+++ b/src/ess/reduce/nexus.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from typing import (
     BinaryIO,
     ContextManager,
+    Literal,
     Mapping,
     NewType,
     Optional,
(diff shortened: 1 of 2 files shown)


src/ess/reduce/nexus.py
```diff
@@ -61,6 +70,7 @@ def load_detector(
     *,
     detector_name: NeXusDetectorName,
     entry_name: Optional[NeXusEntryName] = None,
+    definitions: Optional[Mapping] = None,
```
SimonHeybrock, line 73: Please check, but I think this is *not* the correct default (I hope the unit tests catch this?). Setting the definitions to `None` will override the default definitions, see https://github.com/scipp/scippnexus/blob/1cfab2a69a541832f83d088355eb6fa413a896c8/src/scippnexus/file.py#L15.

jokasimr: Did you check how the argument is used?
See https://github.com/scipp/essreduce/pull/36/files#diff-a0ef1d672f81176717b7e31a2df573d29b3087dd28ce7eb354e1570c78b811b5R298-R300.

Does that still override the default?

SimonHeybrock: ... but then the user cannot set "no definitions"? That may be a valid use case.

jokasimr: Wouldn't that be done by `definitions={}`? Or do you mean that `definitions=None` has a specific meaning in scippnexus?

jokasimr:
> If the file_path argument is an already opened nexus file,
 >   can we add the definitions to that object (requires changes in scippnexus?)
 >   or do we raise an error?

After some discussion about this the conclusion is that we raise an error until there is a concrete use case for this.

src/ess/reduce/nexus.py
SimonHeybrock: In ScippNexus, using `definitions=None` and `definitions={}` is equivalent. It would be confusing if ESSreduce behaved differently (using `definitions=None` to mean "default definitions").

jokasimr: Okay, is there a reason why `scippnexus` doesn't follow the convention that `None` means "use the default value".

Is it correct then that the default value for the `definitions` arguments in essreduce needs to be `snx.base_definitions()`?

jokasimr: Okay, however that causes an issue in this PR. Because if we set the default value of `definitions` to something that is not `None` then we cannot distinguish the case when no definitions were provided from the case when some definitions were provided.

That becomes in issue if the `file_path` is an already opened nexus file (that is, if `file_path` is an `snx.Group`), because if new definitions are provided when we are loading from the group, we have to raise, but if no new definitions were provided we can proceed.

jokasimr: @jl-wynen made the nice suggestion to create a new default value type and use that, seems to solve the problem :+1:

SimonHeybrock (approved):

PR #41, opened 2024-06-20
Replace strict type var naming convention

SimonHeybrock:
Not entirely happy with this either. Should it simply be removed entirely, including the examples in the rule above? Or are we still happy with `RunType`?

### DIFF
--- a/docs/user-guide/reduction-workflow-guidelines.md
+++ b/docs/user-guide/reduction-workflow-guidelines.md
@@ -92,16 +92,13 @@ RunType = TypeVar('RunType', SampleRun, BackgroundRun)
 class Filename(sciline.Scope[RunType, str], str): ...
 ```
 
-### C.3: Carefully choose names of type vars
+### C.3: Use the suffix 'Type' for type vars
 
 **Reason**
-It is important that type vars can be distinguished from concrete domain types, and that it has a clear meaning.
+This makes it easier to distinguish type vars from concrete domain types.
 
 **Example**
-
-- `RunType` and `MonitorType` are examples where having the suffix `Type` can be useful, since `Run` could be confused with (or used as) a domain type giving the run number.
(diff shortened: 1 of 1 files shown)


jl-wynen:
This reads like a general advice but not a hard guideline. So in the interest of making the guidelines enforceable, I'd say remove it.

This advice / guide now reminds me of some naming conventions that encode the kind of a variable in the name. E.g., https://root.cern/TaligentDocs/TaligentOnline/DocumentRoot/1.0/Docs/books/WM/WM_63.html#HEADING77 Personally, I don't like those. We should apply good judgement when picking names.

SimonHeybrock:
Ok, removed. But what about the example for `RunType` and `MonitorType` in the section before? Are we still happy with those recommendations?

jl-wynen (commented):
Can you briefly document the reason for removal? Just so we don't need to dig too much in the future.

SimonHeybrock:
> Can you briefly document the reason for removal? Just so we don't need to dig too much in the future.

Can you suggest something?

docs/user-guide/reduction-workflow-guidelines.md
```diff
@@ -92,13 +92,7 @@ RunType = TypeVar('RunType', SampleRun, BackgroundRun)
...
 ```
 
-### C.3: Use the suffix 'Type' for type vars
-
-**Reason**
-This makes it easier to distinguish type vars from concrete domain types.
-
-**Example**
-See 'RunType' and 'MonitorType' in the table of C.2.
+### C.3: (Removed rule on naming TypeVars)
```
jl-wynen, line 95:
```suggestion
### C.3: (Removed rule on naming TypeVars)

This guideline was too restrictive as not all TypeVars represent a "type", conceptually.
Instead, authors should apply good judgment when naming TypeVars.
```

jl-wynen (approved):

PR #42, opened 2024-06-20
Improve upper bound estimate in broadcast when masks are present

SimonHeybrock:
Fixes scipp/esssans#89.

### DIFF
--- a/src/ess/reduce/uncertainty.py
+++ b/src/ess/reduce/uncertainty.py
@@ -75,7 +75,6 @@ def broadcast_with_upper_bound_variances(
     """
     if _no_variance_broadcast(data, prototype.sizes):
         return data
-    sizes = {dim: size for dim, size in prototype.sizes.items() if dim not in data.dims}
     sizes = prototype.sizes
     mask = sc.scalar(False)
     if isinstance(prototype, sc.DataArray):
--- a/src/ess/reduce/uncertainty.py
+++ b/src/ess/reduce/uncertainty.py
@@ -15,7 +15,9 @@ This module provides two ways of handling variances during broadcast operations:
 from enum import Enum
 from typing import Dict, TypeVar, Union, overload
 
+import numpy as np
 import scipp as sc
+from scipp.core.concepts import irreducible_mask
 
 T = TypeVar("T", bound=Union[sc.Variable, sc.DataArray])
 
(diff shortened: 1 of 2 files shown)


src/ess/reduce/uncertainty.py
```diff
@@ -36,50 +38,99 @@
...
+    Returns
+    -------
+    :
+        The data with the variances scaled by the volume of the new subspace.
+    """
+    if _no_variance_broadcast(data, prototype.sizes):
         return data
-    size = 1
+    sizes = {dim: size for dim, size in prototype.sizes.items() if dim not in data.dims}
+    sizes = prototype.sizes
```
nvaytet, line 79: Is there a typo here? redefining `sizes`?

src/ess/reduce/uncertainty.py
```diff
@@ -36,50 +38,99 @@
...
+        if dim not in data.dims and dim not in mask.dims:
             size *= dim_size
     data = data.copy()
     data.variances *= size
-    return data.broadcast(sizes={**sizes, **data.sizes}).copy()
+    sizes = {**sizes, **data.sizes}
+    data = data.broadcast(sizes=sizes).copy()
+    if mask is not None:
+        # The masked values are not counted in the variance, so we set them to infinity.
+        data.variances[mask.broadcast(sizes=sizes).values] = np.inf
```
nvaytet, line 97: What's the reasoning behind `np.inf`? I know the masked data should not be used, is `np.nan` a better choice? Inf maybe be annoying to handle in plots, fx.

SimonHeybrock: I went by my gut feeling "I am making some error bars smaller, so the others should be larger, to balance it out". Wouldn't scipp/plopp#3292 have solved the issue with plotting you mention?

That being said, if you think NaN actually makes more sense then I can change it.

src/ess/reduce/uncertainty.py
SimonHeybrock: Oops, first line was left over from refactor.

src/ess/reduce/uncertainty.py
nvaytet: Actually, it makes no difference for the plotting, so you can leave it if you like. I though you would get vertical lines spanning the entire height of the figure for the error bars, but in fact nothing is plotted (both my MPL and plotly).
Screenshot at 2024-06-21 10-51-06

I don't know what is a better indicator that the data is invalid there: if it's NaN, does it appear more 'intentional' than inf which could be a result from an operation gone wrong?
In any case, I think it's ok either way.

nvaytet (approved):

PR #46, opened 2024-06-25
feat: add slice argument to nexus loaders

### LINKED ISSUE #45: Support indexing in `load_detector` and `load_monitor`
Currently the `nexus` load functions load everything:
https://github.com/scipp/essreduce/blob/b9a2c28ca04f54e171254eebd693aa44c0537236/src/ess/reduce/nexus.py#L273

We need to support passing a custom index (a dict, for example) here, instead of hard-coding an empty tuple.

This will be essential when processing large files, in particular to load events for a range of pulse-times.

jokasimr:
Fixes #45

### DIFF
--- a/src/ess/reduce/nexus.py
+++ b/src/ess/reduce/nexus.py
@@ -71,7 +71,7 @@ _no_new_definitions = object()
 
 def load_detector(
     file_path: Union[FilePath, NeXusFile, NeXusGroup],
-    data_slice=(),
+    selection=(),
     *,
     detector_name: NeXusDetectorName,
     entry_name: Optional[NeXusEntryName] = None,
@@ -111,7 +111,7 @@ def load_detector(
     return RawDetector(
         _load_group_with_positions(
             file_path,
-            data_slice=data_slice,
+            selection=selection,
             group_name=detector_name,
             nx_class=snx.NXdetector,
             entry_name=entry_name,
@@ -122,7 +122,7 @@ def load_detector(
 
 def load_monitor(
     file_path: Union[FilePath, NeXusFile, NeXusGroup],
-    data_slice=(),
+    selection=(),
     *,
     monitor_name: NeXusMonitorName,
     entry_name: Optional[NeXusEntryName] = None,
@@ -162,7 +162,7 @@ def load_monitor(
     return RawMonitor(
         _load_group_with_positions(
             file_path,
-            data_slice=data_slice,
+            selection=selection,
             group_name=monitor_name,
             nx_class=snx.NXmonitor,
             entry_name=entry_name,
@@ -214,7 +214,7 @@ def load_source(
     return RawSource(
         _load_group_with_positions(
             file_path,
-            data_slice=(),
+            selection=(),
             group_name=source_name,
             nx_class=snx.NXsource,
             entry_name=entry_name,
@@ -266,7 +266,7 @@ def load_sample(
 def _load_group_with_positions(
     file_path: Union[FilePath, NeXusFile, NeXusGroup],
     *,
-    data_slice,
+    selection,
     group_name: Optional[str],
     nx_class: Type[snx.NXobject],
     entry_name: Optional[NeXusEntryName] = None,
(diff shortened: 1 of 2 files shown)


jokasimr:
I'm not so fond of the name `data_slice` for the argument (anything with `data` in the name is a bit shady imo). But `slice` is also not very good because it's a builtin.
Do you have any suggestions?

tests/nexus_test.py
```diff
@@ -289,9 +303,18 @@ def test_load_detector_select_entry_if_not_unique(nexus_file, expected_bank12):
...
+    ],
+)
+def test_load_monitor(nexus_file, expected_monitor, entry_name, data_slice):
+    data_slice = {'data_slice': data_slice} if data_slice is not None else {}
     monitor = nexus.load_monitor(
         nexus_file,
+        **data_slice,
         monitor_name=nexus.NeXusMonitorName('monitor'),
         entry_name=entry_name,
     )
```
SimonHeybrock, line 320: Why is the expected monitor the same in both cases? Looks like we are not testing whether the param is actually used?

tests/nexus_test.py
```diff
@@ -206,9 +206,23 @@ def expected_sample() -> sc.DataGroup:
...
+    ],
+)
+def test_load_detector(nexus_file, expected_bank12, entry_name, data_slice):
+    data_slice = {'data_slice': data_slice} if data_slice is not None else {}
     detector = nexus.load_detector(
         nexus_file,
+        **data_slice,
         detector_name=nexus.NeXusDetectorName('bank12'),
         entry_name=entry_name,
     )
```
SimonHeybrock, line 228: I think a good way to check the loaded data would be to slice during load, and after, and check that the results are the same?

src/ess/reduce/nexus.py
```diff
@@ -71,6 +71,7 @@
 
 def load_detector(
     file_path: Union[FilePath, NeXusFile, NeXusGroup],
+    data_slice=(),
```
SimonHeybrock, line 74: Not happy with the name, but I don't have a better idea right now. @nvaytet @jl-wynen do you?

SimonHeybrock (commented):
Missing update to param lists in docstrings.

tests/nexus_test.py
jokasimr: That's true. I'll fix that

tests/nexus_test.py
jokasimr: Nah this doesn't actually work here. The results are binned by pixel and time and then we can't extract the underlying event coordinates like `event_time_offset`. But when loading from scipp-nexus we can slice by `event_time_offset`.

SimonHeybrock: Why not? Use something like https://scipp.github.io/user-guide/binned-data/filtering.html#Extract-events-matching-parameter-value maybe?

tests/nexus_test.py
SimonHeybrock: Not sure why you think it makes little sense? It should be easy to do?

jokasimr: I just meant we don't really want to test that scippnexus does what it says it will, that is, that slicing the file and slicing the array gives the same result.

The only thing we really want to test here is that the slice is applied to the file.

src/ess/reduce/nexus.py
jokasimr: some alternative names for the argument, kind of in the order I like them most:

* `selection`
* `sel`
* `slice_`
* `nxslice`
* `load_slice`
* `slice_index`
* `index`
* `part`

tests/nexus_test.py
SimonHeybrock: > The only thing we really want to test here is that the provided slice is applied to the file.

Indeed, and how do you check that?

jokasimr: Before I checked it by looking for the dimension error when slicing by a dimension that doesn't exist.
Now I do like you suggested and compare the sliced loaded data and the sliced expected data.

SimonHeybrock: Note that a lot of things in Scipp raise `DimensionError`. Also, it does not tell you if you passed in indices correctly? 👍 for the update.

SimonHeybrock (approved):

PR #52, opened 2024-07-10
Copy JSON NeXus helper from ScippNeutron

jl-wynen:
As discussed in https://github.com/scipp/scippneutron/pull/530


src/ess/reduce/nexus/_nexus_loader.py
YooSunYoung, line 1: Is this file just needed to be moved into the subdirectory?

jl-wynen: I don't understand your question. I moved it because I wanted to group all nexus related functionality into a subpackage. And I don't want to implement anything in `__init__.py`. So I created this file to host the loader functions.

YooSunYoung (approved):

Issue #28, opened 2024-05-16
[Requirement] UI component for selecting, configuring, and running workflows


SimonHeybrock:
### Executive summary

For the "80%" of basic workflow use, we need a simple interface that reduced the need for Python knowledge.

### Context and background knowledge

Users need to create and configure workflows. In many cases the workflow are pre-built in the ESS* packages and need configuration. Simple GUI tools based on (probably?) `ipywidgets` are needed for all ESS* packages. These GUI components will, step by step (e.g., in a series of tabs), allow for:

1. Select a workflow. Assume that we can get a list of workflows from the package, i.e., as input.
2. Set workflow params in widgets.
3. Optionally, define tables of params.
4. Select targets for computation, could be a single final result, or multiple, including intermediate results.
5. Run the workflow.
6. Plot or save outputs.

This requirement is a high-level overview, individual parts will be described in more detail in sub-requirements.

### Inputs

- A package can provide a list of workflows.
- Input params can be obtained from the workflow. There will be a mapping between input keys of the workflow and implementations of a `Param` class. The latter includes things such as a description (to be shown in the GUI), bounds, validators, ...
- Targets could be any key in the pipeline, but likely we will want some customization, i.e., the workflow may specify this.

### Methodology

Python function in a helper module that can be used to auto-generate a widget.

### Outputs

A widget that can be used in a Jupyter notebook. Depending on where we stop in the listing, this might return a configured pipeline, a task graph, results, or even run, plot, and save directly.

### Which interfaces are required?

Python module / function

### Test cases

ESSsans is a good candidate for prototyping and trying this out

SimonHeybrock:
Widget team initial goal:

- What is needed from Python package beyond `Sequence[Workflow]`?
- What does `Workflow` need beyond `Sequence[Parameter]` to generate widgets?
- Find out if draft of `Parameter` class in branch `param-class` has all necessary info.

Parameter/workflow team initial goal:

- How to interface parameters with pipeline. How do we get or define the `Sequence[Parameter]`?

See https://github.com/scipp/esssans/blob/a614f49492453cf6ea0f1615e63898b77b7303f6/src/ess/loki/general.py#L60-L84 for an example of a current workflow, which also defines default params.

YooSunYoung:
> - What is needed from Python package beyond Sequence[Workflow]?

- [ ] Things to be computed - we tried to detect all the output leaves, but then there are providers that are not used.

SimonHeybrock:
[...]

PR #54, opened 2024-08-01
feat: draft autowidget

jokasimr:
Related to #28 

Experiment with `AutoWidget` helper to make writing widgets for parameters a bit easier.

The `AutoWidget` class creates a composite widget from a constructor and a range of argument widgets.
The resulting widget is a simple form that produces a value of the constructor type.


jokasimr:
Not sure how far we can get with this approach or a similar one, but to avoid having to create lots of very similar widgets for different workflow parameters it would be nice to automate what we can

SimonHeybrock:
> but to avoid having to create lots of very similar widgets for different workflow parameters

Can you clarify your concern? I didn't think there would be too many different widgets.

jokasimr:
There's some bookkeeping involved in making a widget that creates a composite value from a number of underlying basic widgets. It's good if we can avoid repeating it unnecessarily.

In particular, the `AutoWidget` let's you observe the composite value, and it's updated when any of the underlying values change. That didn't happen in the old widget implementations. That makes it easier to compose widgets together.

There are other ways to achieve that, not sure if this is the best, but I found it to work quite well so far.

jokasimr:
There's some bookkeeping involved in making a widget that creates a composite value from a number of underlying basic widgets. It's good if we can avoid repeating it unnecessarily.

In particular, the `AutoWidget` let's you observe the composite value, and it's updated when any of the underlying values change. That didn't happen in the old widget implementations. That makes it easier to compose widgets together.

There are other ways to achieve that, not sure if this is the best, but I found it to work quite well so far.

jokasimr:
>  I didn't think there would be too many different widgets.

Not necessarily too many, I agree. I thought there would be more than there seems to be. It might be easier to just keep the widget implementations separate

SimonHeybrock:
If I understood correctly during the standup we will not pursue this for now, may be revisited later?

Issue #61, opened 2024-08-05
Widgets overview


SimonHeybrock:
This is an overview of essentials to resolve for a first public version of the widgets:

- [ ] Implement #59.
- [ ] Implement optional parameters.
- [ ] Implement #57.
- [ ] Show outputs using a vertical layout of checkboxes (instead of list)?
- [ ] Show extended outputs as combo-box, so they are mostly hidden by default
- [ ] Fix some UI logic around enabling/disabling buttons, make sure things can be run again, ...

YooSunYoung:
- [ ] Saving parameters and load them

PR #66, opened 2024-08-08
Add parameter and widget modules

SimonHeybrock:


MridulS (changes requested):
Should we have some tests here?

YooSunYoung (approved):

YooSunYoung:
> Should we have some tests here?

I added that as a task in #61

SimonHeybrock:
I don't mind either way, feel free to either close or merge this, whoever decides to take the lead here!

PR #67, opened 2024-08-08
Switchable widget.

### LINKED ISSUE #57: [UI] Switchable Widget.
Related to #28 

We want a widget that wraps another widget.
If the switch is `on`, it should set the value from the child widget to the workflow, 
if it's `off`, it should not be set at all.

We can either use checkbox or a accordion : D

YooSunYoung:
Fixes #57 

# Preview
https://github.com/user-attachments/assets/d60bf48b-14a6-4ff2-84f5-9b3a9e857c92


src/ess/reduce/widgets/_switchable_widget.py
```diff
@@ -0,0 +1,42 @@
...
+    """Wrapper widget to handle switchable widgets.
+
+    When you retrieve the value of this widget,
+    it will return the value of the wrapped widget.
+    It is expected not to be set in the workflow if ``enabled`` is False.
+    """
+
+    def __init__(self, wrapped: Widget, name: str = '') -> None:
+        super().__init__()
+        self.enable_box = Checkbox(description='Use Parameter', style=default_style)
```
SimonHeybrock, line 20: Can we have no text instead, and grey-out/disable the widget instead? Some widgets (like `BinEdgeWidget`) might be quite wide, so saving space would be good.

src/ess/reduce/widgets/_switchable_widget.py
```diff
@@ -0,0 +1,42 @@
...
+    def enabled(self) -> bool:
+        return self.enable_box.value
+
+    @enabled.setter
+    def enabled(self, value: bool) -> None:
+        self.enable_box.value = value
+
+    @property
+    def value(self) -> Any:
+        return self.wrapped.value
```
SimonHeybrock, line 42: Should have a test to ensure that disabling the widget will disable setting a value on the pipeline.

src/ess/reduce/widgets/_switchable_widget.py
YooSunYoung: If there's no text, it's hard to tell what's the check box is about.

And I couldn't find how to `grey-out` widgets without using java script api... 
And `disable` is not a common property so it's also not an option.

src/ess/reduce/widgets/_switchable_widget.py
YooSunYoung: I added a helper to handle `SwitchableWidget`

https://github.com/scipp/essreduce/blob/024224a4c6e2e9da9ee43b2e13ae02a0961a52b5/src/ess/reduce/ui.py#L115-L127

and used it in the `run_workflow`
https://github.com/scipp/essreduce/blob/024224a4c6e2e9da9ee43b2e13ae02a0961a52b5/src/ess/reduce/ui.py#L130-L147

And then added the test of ``collect_values`` helper instead of building `workflow` for testing.

https://github.com/scipp/essreduce/blob/switchable-widget/tests/widget_test.py

I guess it could also be tested in the higher-level, when we test the whole process.

But it requires registry and simulating selections and everything, so I suggest we do that in another PR.

What do you think...?

src/ess/reduce/widgets/_switchable_widget.py
YooSunYoung: Actually, it's not that bad without the description.

https://github.com/user-attachments/assets/f256927e-49d7-45db-a56d-377d70dc0d73

tests/widget_test.py
```diff
@@ -0,0 +1,39 @@
...
+    assert isinstance(create_parameter_widget(switchable_param), SwitchWidget)
+    non_switchable_param = Parameter('b', 'b', 2, switchable=False)
+    assert not isinstance(create_parameter_widget(non_switchable_param), SwitchWidget)
+
+
+@pytest.mark.filterwarnings(
+    'ignore::DeprecationWarning'
+)  # Ignore deprecation warning from widget library
+def test_collect_values_from_disabled_switchable_widget() -> None:
+    from ess.reduce.ui import collect_values
```
SimonHeybrock, line 22: I think you are kind of testing an implementation detail here: Someone might remove the `collect_values` function, change the mechanism, etc. When that happens, this test will need to be refactored, or worse, it will be removed.

Have you considered testing the actual observable behavior we want, i.e., create a `Pipeline` and check if those values are set by the UI? Is this too complicated to do in the current design? Would something like MVP or MVC make this easier?

tests/widget_test.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+import pytest
+from ess.reduce.parameter import Parameter
+from ess.reduce.widgets import SwitchWidget, create_parameter_widget
+
+
+@pytest.mark.filterwarnings(
+    'ignore::DeprecationWarning'
+)  # Ignore deprecation warning from widget library
```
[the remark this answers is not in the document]
YooSunYoung: I thought it was because of the widget library. But it was because how we set the style. 
I fixed it and removed the mark.

tests/widget_test.py
YooSunYoung: Yes I would rather do that once we have #68  done...?

SimonHeybrock (approved):

PR #70, opened 2024-08-09
Workflow building helper.

### LINKED ISSUE #68: Helper functions to create widgets, instead of creating a single one in the module
We also want a user-guide for how to deploy the widget/app.

YooSunYoung:
Fixes #68

and partially related to #61 

I avoided spending too much time to make it prettier for now.

# Try Out
Now it's building new object everytime you call the helper function.

```python
from ess import loki
from ess.reduce.ui import workflow_widget

workflow_widget()
```

https://github.com/user-attachments/assets/fced63b7-14f9-456c-9d89-44570c578875

There is a smaller helper that create a widget for a certain pipeline constructor.

```python
from ess.reduce.ui import workflow_widget_from_constructor
from ess.loki.general import LokiAtLarmorTutorialWorkflow

workflow_widget_from_constructor(LokiAtLarmorTutorialWorkflow)
```

### DIFF
--- a/src/ess/reduce/ui.py
+++ b/src/ess/reduce/ui.py
@@ -6,7 +6,7 @@ from typing import Any
 import ipywidgets as widgets
 import sciline as sl
 from IPython import display
-from ipywidgets import Layout, TwoByTwoLayout
+from ipywidgets import Layout
 
 from .parameter import Parameter
 from .widgets import SwitchWidget, create_parameter_widget, default_layout
@@ -91,7 +91,9 @@ class ParameterBox(widgets.VBox):
 
         super().__init__([self.parameter_refresh_button, self._input_box], **kwargs)
 
-    def collect_values(self) -> dict[Key, Any]:
+    @property
+    def value(self) -> dict[Key, Any]:
+        """Return the current parameter values with matching types as a dictionary."""
         return {
             node: widget.value
             for node, widget_box in self._input_widgets.items()
@@ -186,7 +188,7 @@ class WorkflowWidget(widgets.TwoByTwoLayout):
         def workflow_runner() -> dict[type, Any]:
             """Run the workflow with the current parameter values."""
             return assign_parameter_values(
-                workflow.copy(), self.parameter_box.collect_values()
+                workflow.copy(), self.parameter_box.value
             ).compute(self.output_selection_box.value)
 
         self.result_box = ResultBox(workflow_runner, result_registry)
@@ -215,7 +217,7 @@ class WorkflowWidget(widgets.TwoByTwoLayout):
 def workflow_widget_from_constructor(
     workflow_constructor: Callable[[], sl.Pipeline],
     result_registry: dict | None = None,
-) -> TwoByTwoLayout:
+) -> WorkflowWidget:
     """Create a widget for a workflow constructed from a workflow constructor."""
     workflow = workflow_constructor()
     return WorkflowWidget(workflow, result_registry)
--- a/src/ess/reduce/ui.py
+++ b/src/ess/reduce/ui.py
@@ -1,14 +1,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
-from collections.abc import Callable, Iterable
-from typing import Any, cast
+from collections.abc import Callable
+from functools import partial
+from typing import Any
 
 import ipywidgets as widgets
 import sciline
 from IPython import display
 from ipywidgets import Layout, TwoByTwoLayout
 
-from .widgets import SwitchWidget, create_parameter_widget, default_style
+from .parameter import Parameter
+from .widgets import SwitchWidget, create_parameter_widget, default_layout
 from .workflow import (
     Key,
     assign_parameter_values,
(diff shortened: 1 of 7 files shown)


YooSunYoung:
@MridulS  I'm gonna add tests and documentation as well but can you please have a look in the meantime?

src/ess/reduce/ui.py
```diff
@@ -18,145 +20,242 @@
...
+class WorkflowWidget(widgets.TwoByTwoLayout):
+    def __init__(
+        self, workflow: sl.Pipeline, result_registry: dict | None = None, **kwargs
+    ):
+        self.output_selection_box = OutputSelectionWidget(workflow)
+        registry_getter = partial(
+            _registry_getter_template, workflow, self.output_selection_box
+        )
+        self.parameter_box = ParameterBox(registry_getter)
+        workflow_runner = partial(
```
jokasimr, line 204: (minor) Since we are not using the `_workflow_runner_template` anywhere else, maybe it's better to create the callback function here in the constructor directly?

Like this:

```python
class WorkflowWidget(widgets.TwoByTwoLayout):
    def __init__(
    ):
        ...

        def workflow_runner():
            # We don't have to copy the workflow here because that's already done in assign_parameter_values
            target_outputs = self.output_selection_box.value
            values = self.input_selection_box.collect_values()
            return assign_parameter_values(workflow, values).compute(
                target_outputs, scheduler=sl.scheduler.NaiveScheduler()
            )
        self.result_box = ResultBox(workflow_runner, result_registry)

            ...
```

The advantage is that it removes one layer of indirection and makes it a bit easier to understand.

Same goes for `_registry_getter_template`.

src/ess/reduce/ui.py
```diff
@@ -18,145 +20,242 @@
...
-    workflow_constructor = cast(Callable[[], sciline.Pipeline], workflow_select.value)
-    selected_workflow = workflow_constructor()
-    outputs = possible_outputs_widget.value + typical_outputs_widget.value
-    registry = get_parameters(selected_workflow, outputs)
 
-    values = collect_values(parameter_box, registry.keys())
+def workflow_widget_from_constructor(
+    workflow_constructor: Callable[[], sl.Pipeline],
+    result_registry: dict | None = None,
+) -> TwoByTwoLayout:
```
jokasimr, line 236: Should the return type just be `WorkflowWidget` now?

src/ess/reduce/ui.py
```diff
@@ -18,145 +20,242 @@
...
+        )
 
-def run_workflow(_: widgets.Button) -> None:
-    workflow_constructor = cast(Callable[[], sciline.Pipeline], workflow_select.value)
-    selected_workflow = workflow_constructor()
-    outputs = possible_outputs_widget.value + typical_outputs_widget.value
-    registry = get_parameters(selected_workflow, outputs)
 
-    values = collect_values(parameter_box, registry.keys())
+def workflow_widget_from_constructor(
```
jokasimr, line 233: Where is this function used? Do we need it?

src/ess/reduce/ui.py
```diff
@@ -18,145 +20,242 @@
...
+                    for node, parameter in registry_getter().items()
+                }
             )
+            self._input_box.children = list(self._input_widgets.values())
+
+        self.parameter_refresh_button.on_click(_refresh_input_box)
+
+        super().__init__([self.parameter_refresh_button, self._input_box], **kwargs)
+
+    def collect_values(self) -> dict[Key, Any]:
```
jokasimr, line 95: (minor) Is there a reason we don't want to call this just `values` instead of `collect_values`?

jokasimr (approved):
Looks nice!

Left some comments, most are smaller things.

src/ess/reduce/ui.py
```diff
@@ -18,145 +20,242 @@
...
+    workflow_constructor: Callable[[], sl.Pipeline],
+    output_selection_box: OutputSelectionWidget,
+    input_selection_box: ParameterBox,
+) -> dict[type, Any]:
+    """Template function that constructs a workflow from inputs of widgets."""
+    workflow = workflow_constructor()
+    target_outputs = output_selection_box.value
+    values = input_selection_box.collect_values()
+    return assign_parameter_values(workflow, values).compute(
+        target_outputs, scheduler=sl.scheduler.NaiveScheduler()
```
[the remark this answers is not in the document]
YooSunYoung: I just copied and pasted.
I guess I can just remove it...?

src/ess/reduce/ui.py
```diff
@@ -18,145 +20,242 @@
...
+        connect_refresh_button(
+            self.parameter_box.parameter_refresh_button, self.result_box.output
+        )
+        connect_output_selection_and_parameter_run_button(
+            self.output_selection_box.typical_outputs_widget,
+            self.output_selection_box.possible_outputs_widget,
+            parameter_refresh_button=self.parameter_box.parameter_refresh_button,
+            run_button=self.result_box.run_button,
+        )
+        for box in (self.output_selection_box, self.parameter_box, self.result_box):
```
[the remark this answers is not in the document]
YooSunYoung: I think it should be fixed/adjusted when we fix the styles overall.
I suggest we keep it here for now to make it easier to adjust the sizeds/locations.

src/ess/reduce/ui.py
YooSunYoung: Yeah, it used to be like this since it didn't have any wrapper class and it was just a ``Box`` instead.
Also I thought maybe each widgets can be used as it is... which is very unlikely I think...
I'll fix it.

src/ess/reduce/ui.py
YooSunYoung: It's for creating a workflow widget from the ``workflow-function`` without selecting it from the higher-widget.

It'll be easier to make a widget quickly from a notebook.

src/ess/reduce/ui.py
YooSunYoung: Just wanted to make it explicit.
But since all the wrapper widget has `value` property, it might be okay to make a `value` property...?

And `values` might be confusing since you might expect it to work like `dict.values`.

src/ess/reduce/ui.py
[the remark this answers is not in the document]
jokasimr: I can't think of a reason why it's needed right now

YooSunYoung:
@jokasimr I either left comments or fixed the code. Can you please have another look?

src/ess/reduce/ui.py
YooSunYoung: I updated it to be a `value` property instead...!

src/ess/reduce/ui.py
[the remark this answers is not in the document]
YooSunYoung: Yes...! I removed it

src/ess/reduce/ui.py
jokasimr: Hm, I don't really understand. It doesn't seem easier to use `workflow_widget_from_constructor(workflow_function, result_registry)` than  `WorkflowWidget(workflow_function(), result_registry)`. But maybe I'm missing the point.

YooSunYoung: Hmm that's true. I think it's also a legacy from the previous cleanup before I made a wrapper classes.
I thought we needed more explicit function since we already have `workflow_widget`...
I'll just delete it for now.

PR #73, opened 2024-08-13
Optional widget wrapper

YooSunYoung:
Fixes scipp/ess#131 

I don't like how it looks but it works...

## Optional

image

## Switchable & Optional

https://github.com/user-attachments/assets/0e26b649-5f4e-461e-acb7-015f65d78065

### DIFF
--- a/docs/developer/gui.ipynb
+++ b/docs/developer/gui.ipynb
@@ -195,6 +195,115 @@
    "source": [
     "ess_widget"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Wrapper Widgets\n",
+    "\n",
+    "In order to handle special cases of parameter settings, we have wrapper widgets.\n",
+    "\n",
+    "Each wrapper widget is associated with certain attribute of ``Parameter`` object.\n",
+    "\n",
+    "They are implemented as a decorator around widget type dispatch function like below.\n",
+    "\n",
+    "It is because of ``@singledispatch`` decorator.\n",
+    "\n",
+    "```python\n",
+    "# In ess.reduce.widgets module\n",
+    "@switchable_widget\n",
+    "@optional_widget  # optional_widget should be applied first\n",
+    "@singledispatch\n",
+    "def create_parameter_widget(param: Parameter) -> widgets.Widget: ...\n",
+    "```\n",
+    "\n",
+    "### Switchable Widget: ``Parameter.switchable``\n",
+    "\n",
+    "Widgets are wrapped in ``SwitchableWidget`` if ``Parameter`` is ``switchable``.\n",
+    "\n",
+    "The wrapped parameter input widget can be turned off and on.\n",
+    "\n",
+    "If the widget is `enabled`(on), the workflow-compute handling widget should set the value as a parameter into the `Pipeline(workflow)`,\n",
+    "but if the widget is not `enabled`(off), the workflow-compute handling widget should skip setting the value as a parameter.\n",
+    "\n",
+    "It means it will either use the default parameter that was set or computed by providers."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from ess.reduce.widgets import create_parameter_widget\n",
+    "from ess.reduce.parameter import Parameter\n",
+    "\n",
+    "switchable_parameter = Parameter(\n",
+    "    name='SwitchableParameter', description=\"\", default=\"\", switchable=True\n",
+    ")\n",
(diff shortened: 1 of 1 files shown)


jokasimr:
I saw this before but didn't understand it. What is a switchable widget? When is it used?

SimonHeybrock:
> I saw this before but didn't understand it. What is a switchable widget? When is it used?

TL;DR: When you have a parameter `param: SomeType | None` (aka `Optional[SomeType]`).

jokasimr:
> TL;DR: When you have a parameter param: SomeType | None (aka Optional[SomeType]).

From @YooSunYoung s explanation that sounds more like an optional widget.

YooSunYoung:
I added the explanation in the developers document

jokasimr:
Instead of making switchable widgets, maybe we should just add a long multiple-select widget containing all domain types (parameters) in the workflow. When you select some of them, they show up in the `ParameterBox` widget and you can assign their values. When you select a set of outputs, the multiple select list of parameters selects the required parameters to produce the outputs.

This would help by

1. removing the need to have switchable parameters/widgets, simplifying the gui
2. make it easy to add support for changing intermediate values, the user just needs to select them manually from the parameter multiple select widget

src/ess/reduce/widgets/_optional_widget.py
```diff
@@ -0,0 +1,66 @@
...
+
+    When you retrieve the value of this widget,
+    it will return the value of the wrapped widget.
+    The parameter should be set as ``None`` or the actual value.
+    """
+
+    def __init__(self, wrapped: Widget, name: str = '', **kwargs) -> None:
+        self.name = name
+        self.wrapped = wrapped
+        self._option_box = RadioButtons(
```
jokasimr, line 21: Maybe it's simpler to use a single checkbox here? That would let us remove the layout code, since we would not have to manually place the two radio elements on the same line.

YooSunYoung:
@jokasimr  I created a separate issue regarding your suggestion.  #74

src/ess/reduce/widgets/_optional_widget.py
jokasimr: I'm thinking something like:

```python
import ipywidgets as ipw

c = ipw.Checkbox(description='')
s = ipw.Stack([ipw.Label('None'), ipw.Text()], selected_index=0)
ipw.dlink((c, 'value'), (s, 'selected_index'))
ipw.HBox([c, s])
```

Screenshot from 2024-08-13 12-49-20
Screenshot from 2024-08-13 12-49-31

src/ess/reduce/widgets/_optional_widget.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        self._option_box.value = None
+        if hasattr(wrapped, "disabled"):
+            # Disable the wrapped widget by default if possible
+            # since the option box is set to None by default
+            wrapped.disabled = True
+
+        # Make the wrapped radio box horizontal
+        self.add_class("widget-optional")
+        _style_html = HTML(
```
jokasimr, line 35: see comment above about checkbox, I think that would let us remove this

src/ess/reduce/widgets/_optional_widget.py
jokasimr: Or, same but using `.disabled` like you did in this PR.

jokasimr: But it doesn't matter much. If you think it's better like it is just go with that.

jokasimr (approved):

src/ess/reduce/widgets/__init__.py
```diff
@@ -56,7 +57,23 @@ def wrapper(param: Parameter) -> widgets.Widget:
...
+        widget = func(param)
+        if param.optional:
+            return OptionalWidget(widget, name=param.name)
+        return widget
+
+    return wrapper
+
+
 @switchable_widget
+@optional_widget  # optional_widget should be applied first
```
jokasimr, line 76: Does this make all widgets optional? Do we want that?

src/ess/reduce/widgets/__init__.py
```diff
@@ -56,7 +57,23 @@ def wrapper(param: Parameter) -> widgets.Widget:
     return wrapper
 
 
+def optional_widget(
```
jokasimr, line 60: Is this decorator meant to be used elsewhere? Right now it's only used in one place as far as I can tell.
Maybe we can just put the logic inside that function for now and make it a decorator later if the logic is reused elsewhere?

YooSunYoung: We can't put it inside because of the singledispatch decorator.
You'll have to do it in every dispatch registered functions.

jokasimr: Aha I see

PR #77, opened 2024-08-20
Fix problems in new NeXus workflows

SimonHeybrock:
This fixes a couple of problems with the new addition from #75, found when rolling this out to ESSsans (see scipp/esssans#160).

### DIFF
--- a/src/ess/reduce/nexus/generic_types.py
+++ b/src/ess/reduce/nexus/generic_types.py
@@ -1,6 +1,7 @@
 """Domain types for use with Sciline, parametrized by run- and monitor-type."""
 
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Generic, NewType, TypeVar
 
 import sciline
@@ -151,8 +152,11 @@ class MonitorData(
 Component = TypeVar('Component', bound=snx.NXobject)
 
 
+class Filename(sciline.Scope[RunType, Path], Path): ...
+
+
 @dataclass
-class Filename(Generic[RunType]):
+class NeXusFileSpec(Generic[RunType]):
     value: FilePath | NeXusFile | NeXusGroup
 
 
--- a/src/ess/reduce/nexus/generic_types.py
+++ b/src/ess/reduce/nexus/generic_types.py
@@ -8,7 +8,7 @@ import sciline
 import scipp as sc
 import scippnexus as snx
 
-from .types import FilePath, NeXusFile, NeXusGroup, NeXusLocationSpec
+from .types import Component, FilePath, NeXusFile, NeXusGroup, NeXusLocationSpec
 
 # 1  TypeVars used to parametrize the generic parts of the workflow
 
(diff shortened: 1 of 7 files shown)


src/ess/reduce/nexus/generic_types.py
```diff
@@ -151,8 +152,11 @@ class MonitorData(
 Component = TypeVar('Component', bound=snx.NXobject)
 
 
+class Filename(sciline.Scope[RunType, Path], Path): ...
+
+
 @dataclass
-class Filename(Generic[RunType]):
+class NeXusFileSpec(Generic[RunType]):
     value: FilePath | NeXusFile | NeXusGroup
```
SimonHeybrock, line 160: @jl-wynen Wondering what you think about this. My latest attempt to have both, Union-type and (generic) type alias, by splitting it into two parts. For the common user there is a helper that forwards `Filename` into `NeXusFileSpec`.

SimonHeybrock: Of course it does not solve the `str | Path` issue, unless we split the types and helpers further.

jl-wynen: I'm not super happy because this adds another node to the graph.

src/ess/reduce/nexus/generic_types.py
```diff
@@ -151,8 +152,11 @@ class MonitorData(
 Component = TypeVar('Component', bound=snx.NXobject)
 
 
+class Filename(sciline.Scope[RunType, Path], Path): ...
```
jl-wynen, line 155: With this, there are now to `Filename` types. The other one is in `types.py`. This seems confusing.

SimonHeybrock: Please see https://github.com/scipp/essreduce/pull/75#discussion_r1719588569, this is the case for about 20 or so types.

src/ess/reduce/nexus/generic_types.py
SimonHeybrock: As you can guess this is due to lack of better options. Do you have any?

SimonHeybrock: We could duplicate the (current 4) providers that turn `NeXusFileSpec` into `NeXusLocationSpec`. Then the extra node is gone. Is that better?

src/ess/reduce/nexus/generic_types.py
jl-wynen: See me new comment

src/ess/reduce/nexus/generic_types.py
jl-wynen: > As you can guess this is due to lack of better options. Do you have any?

Drop `Filename` and auto-convert parameters to their key type on insertion. This would mean that
```python
pl[NeXusFileSpec] = 'some/path'
```
produces the expected `NeXusFileSpec` object.

I suggested this in the past but you decided against it in favour of duck-typing. I think that this precludes us from supporting static type checkers without workarounds like you implemented here.

On a side note, what we really want here is a sum-type that works both at runtime and makes mypy happy. I have been trying to create one several times but have been unsuccessful.

SimonHeybrock: I agree, we should consider your suggestion again.

jl-wynen (approved):

PR #85, opened 2024-08-29
Add streaming module

SimonHeybrock:
This adds `ess.reduce.streaming`, mainly providing `StreamProcessor`. This is intended for use with Beamlime, processing chunks of detector and monitor event-data form a stream.

For context, this is how this could be used with a workflow from ESSsans:

```python
%matplotlib widget
dummy = sc.DataArray(
    sc.zeros(dims=('Q',), shape=(100,), with_variances=True),
    coords={'Q': sc.linspace('Q', 0.0, 1.0, 101, unit='1/angstrom')},
)
fig = dummy.plot(norm='log', scale={'Q': 'log'}, vmin=0.1, vmax=5)
artist = next(iter(fig.artists))
display(fig)
```

```python
from ess.reduce import streaming

streaming_wf = streaming.StreamProcessor(
    base_workflow=workflow,
    dynamic_keys=(
        NeXusMonitorEventData[SampleRun, Incident],
        NeXusDetectorEventData[SampleRun],
    ),
    target_keys=(IofQ[SampleRun],),
    accumulators={
        ReducedQ[SampleRun, Numerator]:streaming.RollingAccumulator(window=5),
        ReducedQ[SampleRun, Denominator]:streaming.RollingAccumulator(window=5),
    },
)

# This Loki data has different det and mon time scales
det_stride = 60
mon_stride = 1
for i in range(100):
    # assume we have loaded events manually somehow, this is not using an actual stream of events
    det_chunk = det_events[det_stride * i : det_stride * (i + 1)].copy()
    mon_chunk = mon_events[mon_stride * i : mon_stride * (i + 1)].copy()
    # simulate temporary loss of detector data
    if 20 < i < 30:
        det_chunk *= 0.0
    results = streaming_wf.add_chunk(
        {
            NeXusDetectorEventData[SampleRun]: det_chunk,
            NeXusMonitorEventData[SampleRun, Incident]: mon_chunk,
        }
    )
    fig.update({artist: results[IofQ[SampleRun]]})
    fig.fig.canvas.draw()
    fig.fig.canvas.flush_events()
```

### DIFF
--- a/src/ess/reduce/streaming.py
+++ b/src/ess/reduce/streaming.py
@@ -141,10 +141,11 @@ class StreamProcessor:
     def __init__(
         self,
         base_workflow: sciline.Pipeline,
+        *,
         dynamic_keys: tuple[sciline.typing.Key, ...],
-        accumulation_keys: tuple[sciline.typing.Key, ...],
         target_keys: tuple[sciline.typing.Key, ...],
-        accumulator: type[Accumulator] = EternalAccumulator,
+        accumulators: dict[sciline.typing.Key, Accumulator]
+        | tuple[sciline.typing.Key, ...],
     ) -> None:
         """
         Create a stream processor.
@@ -155,12 +156,12 @@ class StreamProcessor:
             Workflow to be used for processing chunks.
         dynamic_keys:
             Keys that are expected to be updated with each chunk.
-        accumulation_keys:
-            Keys for which to accumulate values.
         target_keys:
             Keys to be computed and returned.
-        accumulator:
-            Accumulator class to use for accumulation.
+        accumulators:
+            Keys at which to accumulate values and their accumulators. If a tuple is
+            passed, :py:class:`EternalAccumulator` is used for all keys. Otherwise, a
+            dict mapping keys to accumulator instances can be passed.
         """
         workflow = sciline.Pipeline()
         for key in target_keys:
@@ -177,8 +178,11 @@ class StreamProcessor:
 
         self._process_chunk_workflow = workflow.copy()
         self._finalize_workflow = workflow.copy()
-        # TODO: We may need to have a way to specify the accumulator for each key
-        self._accumulators = {key: accumulator() for key in accumulation_keys}
+        self._accumulators = (
+            accumulators
+            if isinstance(accumulators, dict)
+            else {key: EternalAccumulator() for key in accumulators}
+        )
         self._target_keys = target_keys
 
     def add_chunk(
--- a/src/ess/reduce/streaming.py
+++ b/src/ess/reduce/streaming.py
@@ -128,8 +128,15 @@ class RollingAccumulator(Accumulator[T]):
             self._values.pop(0)
 
 
-class Streaming:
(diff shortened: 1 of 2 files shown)


tests/streaming_test.py
jl-wynen, line 1: Missing a test of `StreamProcessor` with a different accumulator. (To ensure that the accumulator is actually used as expected.)

src/ess/reduce/streaming.py
```diff
@@ -0,0 +1,213 @@
...
+    the accumulation keys.
+    """
+
+    def __init__(
+        self,
+        base_workflow: sciline.Pipeline,
+        dynamic_keys: tuple[sciline.typing.Key, ...],
+        accumulation_keys: tuple[sciline.typing.Key, ...],
+        target_keys: tuple[sciline.typing.Key, ...],
+        accumulator: type[Accumulator] = EternalAccumulator,
```
jl-wynen, line 147: Doesn't this need to allow for different accumulator for each key? That is, should this be `dict[sciline.typing.Key, type[Accumulator]]`, where the keys are the same as in `accumulation_keys`? Alternatively, the `accumulation_keys` and `accumulator` args could be merged.

Found the todo comment after writing this. But it still stands.

src/ess/reduce/streaming.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # Find and pre-compute static nodes as far down the graph as possible
+        # See also https://github.com/scipp/sciline/issues/148.
+        nodes = _find_descendants(workflow, dynamic_keys)
+        parents = _find_parents(workflow, nodes) - nodes
+        for key, value in base_workflow.compute(parents).items():
+            workflow[key] = value
+
+        self._process_chunk_workflow = workflow.copy()
+        self._finalize_workflow = workflow.copy()
```
jl-wynen, line 179: Is there a simple way to split the graph at the accumulated nodes? It took me a moment to understand why there are two copies of the same workflow. Plus, splitting would reduce the risk of misusing the workflows.

src/ess/reduce/streaming.py
SimonHeybrock: I don't think it is worth changing this before we have more data on how it will be used. We may be missing something else entirely.

src/ess/reduce/streaming.py
SimonHeybrock: This does look very simple: https://stackoverflow.com/questions/65853641/networkx-how-to-split-nodes-into-two-effectively-disconnecting-edges

I have previously considered extending the slicing syntax to allow for specifying start and end using `graph[A:B]`, but that is not implemented yet. Also, I would not work if multiple starts/ends are required (unless we want to support a list syntax, `graph[A:[B1,B2]]`).

src/ess/reduce/streaming.py
SimonHeybrock: Well, I changed my mind when trying to write some more tests...

jl-wynen (approved):

PR #86, opened 2024-09-02
Support preopening NeXus files

SimonHeybrock:
This is not strictly required by anything right now, but I needed it for some testing, so a cleaned it up for the future.

Changes:
- Be more relaxed about passing definitions for open groups.
- Add flag to optionally open NeXus files. This may become more relevant in the future when loading even more different parts and/or chunks from files.

### DIFF
--- a/src/ess/reduce/nexus/generic_workflow.py
+++ b/src/ess/reduce/nexus/generic_workflow.py
@@ -17,15 +17,15 @@ import scippnexus as snx
 from . import generic_types as gt
 from . import workflow
 from .generic_types import MonitorType, PulseSelection, RunType
-from .types import (
-    DetectorBankSizes,
-    GravityVector,
-    NeXusDetectorName,
-)
+from .types import DetectorBankSizes, GravityVector, NeXusDetectorName, PreopenNeXusFile
 
 
-def file_path_to_file_spec(filename: gt.Filename[RunType]) -> gt.NeXusFileSpec[RunType]:
-    return gt.NeXusFileSpec[RunType](filename)
+def file_path_to_file_spec(
+    filename: gt.Filename[RunType], preopen: PreopenNeXusFile
+) -> gt.NeXusFileSpec[RunType]:
+    return gt.NeXusFileSpec[RunType](
+        snx.File(filename, definitions=workflow.definitions) if preopen else filename
+    )
 
 
 def no_monitor_position_offset() -> gt.MonitorPositionOffset[RunType, MonitorType]:
@@ -238,16 +238,25 @@ _detector_providers = (
     assemble_detector_data,
 )
 
+_default_params = {
+    PulseSelection: PulseSelection(()),
+    PreopenNeXusFile: PreopenNeXusFile(False),
+}
+
 
 def LoadMonitorWorkflow() -> sciline.Pipeline:
     """Generic workflow for loading monitor data from a NeXus file."""
-    wf = sciline.Pipeline((*_common_providers, *_monitor_providers))
+    wf = sciline.Pipeline(
+        (*_common_providers, *_monitor_providers), params=_default_params
+    )
     return wf
 
 
 def LoadDetectorWorkflow() -> sciline.Pipeline:
     """Generic workflow for loading detector data from a NeXus file."""
-    wf = sciline.Pipeline((*_common_providers, *_detector_providers))
+    wf = sciline.Pipeline(
+        (*_common_providers, *_detector_providers), params=_default_params
+    )
     wf[DetectorBankSizes] = DetectorBankSizes({})
     return wf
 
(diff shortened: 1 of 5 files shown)


src/ess/reduce/nexus/generic_workflow.py
```diff
@@ -17,15 +17,15 @@
...
 
 
-def file_path_to_file_spec(filename: gt.Filename[RunType]) -> gt.NeXusFileSpec[RunType]:
-    return gt.NeXusFileSpec[RunType](filename)
+def file_path_to_file_spec(
+    filename: gt.Filename[RunType], preopen: PreopenNeXusFile
+) -> gt.NeXusFileSpec[RunType]:
+    return gt.NeXusFileSpec[RunType](
+        snx.File(filename, definitions=workflow.definitions) if preopen else filename
+    )
```
jl-wynen, line 28: I don't like this. Normally, we would open the file and pass the handle to a loader. I would expect the same from a workflow. That is, I would want to do this:
```python
handle = snx.File(...)
params[gt.Filename[SampleRun]] = handle
```

I thought the duck-typing approach would allow this?

src/ess/reduce/nexus/workflow.py
```diff
@@ -475,6 +471,11 @@ def _add_variances(da: sc.DataArray) -> sc.DataArray:
     return out
 
 
+definitions = snx.base_definitions()
+definitions["NXdetector"] = _StrippedDetector
+definitions["NXmonitor"] = _StrippedMonitor
```
jl-wynen, line 476: Can you use a function or an immutable dict instead? I don't like mutable global (public) variables.

src/ess/reduce/nexus/generic_workflow.py
```diff
@@ -220,24 +220,34 @@ def assemble_monitor_data(
     assemble_detector_data,
 )
 
+_default_params = {
+    PulseSelection: PulseSelection(()),
```
jl-wynen, line 224: Why is this needed now? It seems unrelated to the other changes.

SimonHeybrock: Avoids 3x code duplication?

src/ess/reduce/nexus/generic_workflow.py
SimonHeybrock: Well, that would not pass mypy, since `Filename` cannot be a file handle?

jl-wynen: Isn't this the whole point of not checking the types of parameters any more?

You are right that, in principle, it shouldn't pass. But it does pass.

src/ess/reduce/nexus/generic_workflow.py
SimonHeybrock: You are right, that was a bad rebase, fixing.

src/ess/reduce/nexus/generic_workflow.py
SimonHeybrock: I am not sure I understand what you are saying. There appear to be 3 different things you mention: Having a provider open a file, mypy, and not checking types of parameters in Sciline.

They all seem kind of distinct. We should have a chat so you can explain!

I would like to control the behavior via a parameter, since having to change a ton of user code in a very complicated manner does not seem feasible or remotely user friendly.

jl-wynen:
> @jl-wynen is this related to [scipp/scippnexus#238](https://github.com/scipp/scippnexus/pull/238) ?

Yes. But ESSreduce has its own mechanism for this. See https://github.com/scipp/essreduce/blob/dd87cd7264f0187dd888ce6afcd7096df2ca5391/src/ess/reduce/nexus/_nexus_loader.py#L65

SimonHeybrock:
As mentioned in the discussion related to (not) locking (#99) earlier, I think doing this via an injected provider seems like a decent option to me. The question is whether having what we have in this PR as a default is reasonable, or if should be something else.

As this sees little use for now, we can also consider to go with it, then iterate.

Please review again under those considerations!

src/ess/reduce/nexus/workflow.py
jl-wynen: This comment still applies.

jl-wynen:
> As mentioned in the discussion related to (not) locking (https://github.com/scipp/essreduce/issues/99) earlier, I think doing this via an injected provider seems like a decent option to me.

What do you mean by 'injected provider'? I don't see one in this PR. So are you saying that we should do it differently?

SimonHeybrock:
> > As mentioned in the discussion related to (not) locking (#99) earlier, I think doing this via an injected provider seems like a decent option to me.
> 
> What do you mean by 'injected provider'? I don't see one in this PR. So are you saying that we should do it differently?

This PR (re)defines `file_path_to_file_spec`. What I am saying is that if a user requires special features such as non-locking or SWMR mode they can insert their own into the workflow, kicking out the pre-defined one.

jl-wynen (approved):

PR #88, opened 2024-09-02
Make PulseSelection generic

### LINKED ISSUE #87: `PulseSelection` must be separate for each run type
https://github.com/scipp/essreduce/blob/168d2f436d947d4286b9d5186dde72096f567216/src/ess/reduce/nexus/types.py#L72

Selecting the same range of pulse across different run types almost never makes sense. Rename this to `AnyRunPulseSelection`, add `PulseSelection[RunType]` in `generic_types.py`.

SimonHeybrock:
Fixes #87.

### DIFF
--- a/src/ess/reduce/nexus/generic_types.py
+++ b/src/ess/reduce/nexus/generic_types.py
@@ -157,6 +157,13 @@ class Filename(sciline.Scope[RunType, Path], Path): ...
 
 
 @dataclass
+class PulseSelection(Generic[RunType]):
+    """Range of neutron pulses to load from NXevent_data groups."""
+
+    value: slice
+
+
+@dataclass
 class NeXusFileSpec(Generic[RunType]):
     value: FilePath | NeXusFile | NeXusGroup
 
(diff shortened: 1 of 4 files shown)


src/ess/reduce/nexus/generic_types.py
```diff
@@ -156,6 +156,13 @@ class MonitorData(
 class Filename(sciline.Scope[RunType, Path], Path): ...
 
 
+@dataclass
+class PulseSelection(Generic[RunType]):
+    """Range of neutron pulses to load from NXevent_data groups."""
+
+    value: slice
```
jl-wynen, line 163: Is this the correct type? The code in the PR passes an empty tuple. So should this be something like ScipnnNexus' indexing type?

src/ess/reduce/nexus/types.py
```diff
@@ -69,7 +69,7 @@
 AnyRunDetectorData = NewType('AnyRunDetectorData', sc.DataArray)
 AnyRunAnyMonitorData = NewType('AnyRunAnyMonitorData', sc.DataArray)
 
-PulseSelection = NewType('PulseSelection', slice)
+AnyRunPulseSelection = NewType('AnyRunPulseSelection', slice)
```
jl-wynen, line 72: Same as `PulseSelection`.

src/ess/reduce/nexus/generic_types.py
SimonHeybrock: Good catch, I changed code to pass a slice. I am not sure why an empty slice even worked with ScippNexus, since the value `PulseSelection` is put into a `dict` for the value of `event_time_zero`.

PR #95, opened 2024-09-04
Fix signal finding in Nexus group tripping up on time-dependent transform

SimonHeybrock:


jokasimr (approved):

jokasimr:
Might there be other names we want to ignore? Is there a way we can figure out what they are or should we just wait and see until we encounter them?

SimonHeybrock:
Should mostly be listed in https://manual.nexusformat.org/classes/base_classes/NXdetector.html, but NXcite was not (but NXcite says it can be on an NXdetector...). I think we should just wait.

SimonHeybrock:
> Might there be other names we want to ignore? Is there a way we can figure out what they are or should we just wait and see until we encounter them?

Hard to say. These two are added by ScippNexus. If someone put, e.g., NXlog in the NXdetector it will break again, but then it is actually not a false positive.

PR #104, opened 2024-09-23
Load histogram monitors

jl-wynen:
This adds the option to load histogram data for monitors instead of event data. While we may not need this in production, simulated data may use histogram monitors instead of event monitors.

In particular, ESSspectroscopy needs this feature. I tested it locally with that pacakge. Note #103 about testing.

Issue #103, opened 2024-09-23
Test NeXus workflow


jl-wynen:
`LoadNeXusWorkflow` is currently untested. We should change that!

SimonHeybrock:
Most of the components are tested in https://github.com/scipp/essreduce/blob/main/tests/nexus/generic_workflow_test.py and https://github.com/scipp/essreduce/blob/main/tests/nexus/workflow_test.py. `LoadNeXusWorkflow` itself is a helper that is currently not (and may never) be used by anything. It is meant as a quick helper to work with unknown files, so it is unclear what its future will be.

PR #106, opened 2024-09-24
Load histogram data

jl-wynen:
Alternative to #104. This PR makes it so that the nexus workflow can automatically load histogram data if a file contains that instead of events.

### DIFF
--- a/src/ess/reduce/nexus/_nexus_loader.py
+++ b/src/ess/reduce/nexus/_nexus_loader.py
@@ -13,7 +13,13 @@ import scipp as sc
 import scippnexus as snx
 
 from ..logging import get_logger
-from .types import AnyRunFilename, NeXusEntryName, NeXusGroup, NeXusLocationSpec
+from .types import (
+    AnyRunFilename,
+    AnyRunPulseSelection,
+    NeXusEntryName,
+    NeXusGroup,
+    NeXusLocationSpec,
+)
 
 
 class NoNewDefinitionsType: ...
@@ -163,9 +169,23 @@ def _select_unique_array(
     return next(iter(arrays.values()))
 
 
+def _to_snx_selection(
+    selection: snx.typing.ScippIndex | AnyRunPulseSelection, *, for_events: bool
+) -> snx.typing.ScippIndex:
+    match selection:
+        case AnyRunPulseSelection(slice(start=None, stop=None)):
+            return ()
+        case AnyRunPulseSelection(sel):
+            if for_events:
+                return {'event_time_zero': sel}
+            return {'time': sel}
+        case _:
+            return selection
+
+
 def load_data(
     file_path: AnyRunFilename,
-    selection: snx.typing.ScippIndex = (),
+    selection: snx.typing.ScippIndex | AnyRunPulseSelection = (),
     *,
     entry_name: NeXusEntryName | None = None,
     component_name: str,
@@ -213,15 +233,17 @@ def load_data(
         component = instrument[component_name]
         if _contains_nx_class(component, snx.NXevent_data):
             data = _unique_child_group(component, snx.NXevent_data, None)
+            sel = _to_snx_selection(selection, for_events=True)
         elif _contains_nx_class(component, snx.NXdata):
             data = _unique_child_group(component, snx.NXdata, None)
+            sel = _to_snx_selection(selection, for_events=False)
         else:
             raise ValueError(
                 f"NeXus group '{component.name}' contains neither "
                 "NXevent_data nor NXdata."
             )
 
-        return data[selection]
+        return data[sel]
 
 
 def group_event_data(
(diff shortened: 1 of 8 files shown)


src/ess/reduce/nexus/_nexus_loader.py
```diff
@@ -180,14 +222,25 @@ def load_event_data(
...
         instrument = _unique_child_group(entry, snx.NXinstrument, None)
         component = instrument[component_name]
-        event_data = _unique_child_group(component, snx.NXevent_data, None)
-        return event_data[selection]
+        if _contains_nx_class(component, snx.NXevent_data):
+            data = _unique_child_group(component, snx.NXevent_data, None)
+            sel = _to_snx_selection(selection, for_events=True)
+        elif _contains_nx_class(component, snx.NXdata):
+            data = _unique_child_group(component, snx.NXdata, None)
+            sel = _to_snx_selection(selection, for_events=False)
```
SimonHeybrock, line 236: If the monitor or detector is not time-dependent then there will not be an `NXdata`, just a HDF5 dataset.

src/ess/reduce/nexus/workflow.py
```diff
@@ -103,10 +103,10 @@ def monitor_events_by_name(
     name:
         Name of the monitor group.
     selection:
-        Selection (start and stop as a Python slice object) for the monitor event data.
+        Selection (start and stop as a Python slice object) for the monitor data.
     """
-    return NeXusEventDataLocationSpec[snx.NXmonitor](
-        filename=filename, component_name=name, selection={'event_time_zero': selection}
+    return NeXusDataLocationSpec[snx.NXmonitor](
+        filename=filename, component_name=name, selection=selection.as_snx_selection()
```
SimonHeybrock, line 109: Can't see where this method is defined?

jl-wynen: Leftover from a previous attempt.

src/ess/reduce/nexus/_nexus_loader.py
jl-wynen: How do we identify the data in that case?

SimonHeybrock: ScippNexus does this, in can be a bit complicated in general, see https://github.com/scipp/scippnexus/blob/94779e6d5e10666eaccc7d343079f9762ef9e3ea/src/scippnexus/nxdata.py#L63-L69 as a starting point.

SimonHeybrock: This explains most of it: https://github.com/scipp/scippnexus/blob/94779e6d5e10666eaccc7d343079f9762ef9e3ea/src/scippnexus/nxdata.py#L116-L123

jl-wynen: So how about this?
```python
if _contains_nx_class(component, snx.NXevent_data):
    data = _unique_child_group(component, snx.NXevent_data, None)
    sel = _to_snx_selection(selection, for_events=True)
elif _contains_nx_class(component, snx.NXdata):
    data = _unique_child_group(component, snx.NXdata, None)
    sel = _to_snx_selection(selection, for_events=False)
elif _is_subclass_of_nxdata(component):
    data = component
    sel = _to_snx_selection(selection, for_events=False)
else:
    raise ValueError(
        f"NeXus group '{component.name}' contains neither "
        "NXevent_data nor NXdata."
    )
```
Or is `component` not required to be something compatible with NXdata?

SimonHeybrock: I think that might lead to loading the entire NXdetector nested, a second time?

jl-wynen: Depends on what `component` is. But yes, this can happen.

Can we just not support this case for now. The current implementation works for BIFROST. And we can always extend it later.

SimonHeybrock:
Have you tested if this still passes all the ESSsans and ESSdiffraction tests/builds?

jl-wynen:
No. I was not aware that those count as tests for ESSreduce. I can do this now. But we very much should implement #103

SimonHeybrock:
> No. I was not aware that those count as tests for ESSreduce. I can do this now. But we very much should implement #103

These are not using `LoadNexusWorkflow`, but `GenericNexusWorkflow`. There are tests in https://github.com/scipp/essreduce/blob/main/tests/nexus/workflow_test.py and https://github.com/scipp/essreduce/blob/main/tests/nexus/generic_workflow_test.py, but as this is a rather complex piece I'd still prefer to run those downstream checks.

jl-wynen:
Those tests don't test the workflows, only individual providers.

SimonHeybrock:
They do? https://github.com/scipp/essreduce/blob/aef4a3c1094dd15f1849aedc74e9960e1b922de3/tests/nexus/generic_workflow_test.py#L36-L51

jl-wynen:
All tests and docs in essdiffraction and esssans pass.

SimonHeybrock (approved):

PR #107, opened 2024-09-26
fix: remove unnecessary column

### LINKED ISSUE #100: Duplicate detector number?
It seems like we end up with two identical coords when loading nexus files: `detector_number` and `event_id`. E.g., in the LoKI direct beam workflow:
Screenshot_20240919_152256

Can we skip one of them? That would save some memory and maybe run time. (The arrays are not aliases of each other.)

jokasimr:
Fixes #100

### DIFF
--- a/src/ess/reduce/nexus/_nexus_loader.py
+++ b/src/ess/reduce/nexus/_nexus_loader.py
@@ -286,7 +286,9 @@ def group_event_data(
         or (begin[1:] != end[:-1]).any()
     ):
         raise ValueError("Grouping only implemented for contiguous data with no masks.")
-    return data.group(event_id).fold(dim='event_id', sizes=detector_number.sizes)
+    out = data.group(event_id).fold(dim='event_id', sizes=detector_number.sizes)
+    out.coords['detector_number'] = out.coords.pop('event_id')
+    return out
 
 
 def _format_time(time: sc.Variable | None) -> str:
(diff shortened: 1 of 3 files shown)


src/ess/reduce/nexus/workflow.py
```diff
@@ -389,6 +389,7 @@ def assemble_detector_data(
     grouped = nexus.group_event_data(
         event_data=event_data, detector_number=detector.coords['detector_number']
     )
+    grouped.coords.pop('event_id')
```
jl-wynen, line 392: I think this should be done in `group_event_data` because that function adds `event_id` as an auxiliary coord. So it should never escape that function.

jl-wynen (changes requested):

src/ess/reduce/nexus/workflow.py
jokasimr: I was thinking the opposite. If it is removed, how to tell what `event_id` the events in each bin are associated with? Removing the coord makes the return value of `group_event_data` useless unless the caller is knowledgeable enough to be aware that they need to re-add the `detector_number` coord.

SimonHeybrock: I think we should rename `event_id` to `detector_number` before returning from `group_event_data`.

jl-wynen (approved):

PR #109, opened 2024-09-27
Binedges widget

### LINKED ISSUE #59: `BinEdgeWidget` and improved `BinEdgeParam`
Goal: Create a widget for defining bin edges from a given workflow parameter, based on and instance of `BinEdgeParam`.

The widget should be more compact than the current draft, i.e., everything in a single line.

Features:
- User can select between linspace and geomspace.
- Define start, stop, and num
- `BinEdgeParam` must be able to define defaults for the three required inputs.

Optional features (we may choose to ignore this initially, it might never be required):
- Select unit from a pre-defined set.
- Consider if we want/need bounds for the params. `num<2` never makes sense so that can be automatic.

nvaytet:
Make a widget for bin edges.
Fixes #59 

Examples:
```Py
# By default, unit is deduced from dim
parameter_registry[WavelengthBins] = BinEdgesParameter(
    WavelengthBins, dim='wavelength'
)
```
Screenshot at 2024-09-27 15-24-07

```Py
# Overriding the unit
parameter_registry[WavelengthBins] = BinEdgesParameter(
    WavelengthBins, dim='wavelength', unit='m'
)
```
Screenshot at 2024-09-27 15-26-17

```Py
# Supplying default values
parameter_registry[WavelengthBins] = BinEdgesParameter(
    WavelengthBins, dim='wavelength', start=0.01, stop=12.0, nbins=300, log=True
)
```
Screenshot at 2024-09-27 15-27-34

### DIFF
--- a/src/ess/reduce/widgets/_binedges_widget.py
+++ b/src/ess/reduce/widgets/_binedges_widget.py
@@ -48,8 +48,8 @@ class BinEdgesWidget(ipw.HBox, ipw.ValueWidget):
             "start": ipw.FloatText(description='start:', value=start, **style),
             "stop": ipw.FloatText(description='stop:', value=stop, **style),
             "nbins": ipw.BoundedIntText(
-                description='nbins:', value=nbins, min=1, **style
-            ),
+                description='nbins:', value=nbins, min=1, max=int(1e9), **style
+            ),  # Note that a max value is required as it defaults to 100 without it.
             "spacing": ipw.Dropdown(
                 options=['linear', 'log'],
                 value='log' if log else 'linear',
--- a/src/ess/reduce/widgets/_binedges_widget.py
+++ b/src/ess/reduce/widgets/_binedges_widget.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
 import ipywidgets as ipw
-import numpy as np
 import scipp as sc
 
 UNITS_LIBRARY = {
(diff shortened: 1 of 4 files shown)


SimonHeybrock:
In your screenshots you are creating the widget for `WavelengthBins`, but the widget shows "Binning: wavelength". Shouldn't it show the name of the domain-type?

Does overriding the unit allow for passing a sequence of possible units?

src/ess/reduce/widgets/_binedges_widget.py
```diff
@@ -0,0 +1,79 @@
...
+            "start": self.fields['start'].value,
+            "stop": self.fields['stop'].value,
+        }
+
+        if self.fields["spacing"].value == "linear":
+            return sc.linspace(**kwargs)
+        else:
+            for key in ("start", "stop"):
+                kwargs[key] = np.log10(kwargs[key])
+            return sc.logspace(**kwargs)
```
SimonHeybrock, line 79: Why not `geomspace`?

SimonHeybrock (commented):
Have you tried updating the ESSsans branch? Does it work nicely there and fit into the layout?

src/ess/reduce/widgets/_binedges_widget.py
nvaytet: hmm yes, that's better, thanks

nvaytet:
> Have you tried updating the ESSsans branch? Does it work nicely there and fit into the layout?

Yes, see https://github.com/scipp/esssans/pull/171
In the notebook, it looks like this
Screenshot at 2024-09-30 09-43-36

SimonHeybrock (approved):

nvaytet:
We should probably improve the look of the vector widget in a similar way ;-)

PR #111, opened 2024-09-30
Better looking vector widget

nvaytet:
Before:
Screenshot at 2024-09-30 10-06-08

After:
Screenshot at 2024-09-30 10-17-52

### DIFF
--- a/src/ess/reduce/parameter.py
+++ b/src/ess/reduce/parameter.py
@@ -146,8 +146,13 @@ class ScalarParamWithUnitOptions(Parameter[T]):
 
 
 @dataclass(kw_only=True)
-class VectorParameter(Parameter[sc.Variable]):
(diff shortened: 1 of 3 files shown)


MridulS (approved):

SimonHeybrock:
Isn't `z` actually unused, i.e., we should not use a generic vector widget?

nvaytet:
> Isn't `z` actually unused, i.e., we should not use a generic vector widget?

So should we make a `Vector2dWidget` and a `Vector3dWidget`?

SimonHeybrock:
> > Isn't `z` actually unused, i.e., we should not use a generic vector widget?
> 
> So should we make a `Vector2dWidget` and a `Vector3dWidget`?

Maybe?

SimonHeybrock (approved):

PR #112, opened 2024-10-01
Remove non-generic nexus components

### LINKED ISSUE #103: Test NeXus workflow

SimonHeybrock:
These were added initially when it was not clear that we would go with a fully generic workflow, and left in since I thought they might be useful for debugging or general exploratory work. However, it now proves that they simply add friction in further code changes, so there is no benefit.

Fixes #103 by removing it.

I have checked on ESSsans that downstream packages will simply have to update a few imports, i.e., updating this will be trivial.

There are no code changes here, apart form removing forwarding and updating type-hints.


src/ess/reduce/nexus/types.py
```diff
@@ -25,80 +26,223 @@
...
+    value: snx.typing.ScippIndex | slice
+
+    def to_snx_selection(self, *, for_events: bool) -> snx.typing.ScippIndex:
+        if self.value == slice(None, None):
+            return ()
+        if isinstance(self.value, slice):
+            if for_events:
+                return {'event_time_zero': self.value}
+            return {'time': self.value}
+        return self.value
```
jl-wynen, line 201: Duplicate of `_nexus_loader._to_snx_selection`. Why did you change the loader to take a raw selection instead of a `PulseSelection` object?

src/ess/reduce/nexus/workflow.py
```diff
@@ -509,110 +552,58 @@ def _add_variances(da: sc.DataArray) -> sc.DataArray:
...
-        Length of chunks as a time difference, i.e., with a time unit.
-    """
-    import pandas as pd
-
-    info = wf.compute(nexus.NeXusFileInfo)
-    # start_time and end_time are taken from event_time_zero, so we always want to
-    # include the end
-    bounds = sc.arange(
-        'chunk', info.start_time, info.end_time + chunk_length, chunk_length
+def GenericNeXusWorkflow() -> sciline.Pipeline:
```
jl-wynen, line 602:
```suggestion
def LoadNeXusWorkflow() -> sciline.Pipeline:
```
Since you removed 'generic' everywhere else.

src/ess/reduce/nexus/types.py
SimonHeybrock: Because `PulseSelection` is now generic, so that didn't work.

Thought I removed this (I was moving it back and forth), good catch.

src/ess/reduce/nexus/workflow.py
SimonHeybrock: "Generic" here means "cross-technique".

jl-wynen: You didn't use that word for the other workflows in this file. And they are also cross-technique

SimonHeybrock: I don't see the point in changing naming in this context. If you think it is really important we can sit down and have a session on overall naming conventions.

jl-wynen: I don't think it is very important. I just think that 'GenericNeXusWorkflow' sounds like it does something conceptually different than 'LoadMonitorWorkflow'. Where actually, it just loads data and is as generic as 'LoadMonitorWorkflow'.

jl-wynen (approved):
Naming is up to you

Issue #96, opened 2024-09-05
Nexus workflow module: Handling time-dependent NXtransformations


SimonHeybrock:
When component positions in a NeXus file are not fixed, this is represented as one or more transformations with a value dataset replaced by a time-series (NXlog). In this case it is not possible to compute, e.g., a unique `sample_position` or `position` coordinate when loading.

The current workflows simply fail, since `scippnexus.compute_positions` will not create a detector position coordinate &mdash; instead the surrounding `DataGroup` will have a time-dependent `position` `DataArray`. This is deliberately not multiplied into the pixel position offsets since this could and would lead to a massive memory use (in addition to leading to a coordinate with time-dependence, which is not allowed by Scipp unless the data has this coord, which it does not).

In practice, I think data reduction will have to be performed for various time intervals, with constant positions in each interval. We somehow need to facilitate handling this, extracting relevant information from various parts. This included determining a unique position coordinate, as well as loading the correct chunk of events. The current `GenericNeXusWorkflow` and related components do not support this. We need to figure out a convenient and safe mechanism to make this work. There are a bunch of cases:

1. Detector moves, e.g., to cover different angles.
2. Detector moves to different distance.
3. Sample rotates (SXD).

For now, I think we should focus on the case of step-wise movement but keep the continuous scanning case in the back of our minds.

SimonHeybrock:
An easy subgoal that will likely cover many uses cases is to improve the loaders to take into account the time interval. Currently `ess.reduce.nexus` supports a `PulseSelection` but this is ignored when loading any `NXlog`. I propose to change (or extend) `PulseSelection` to a more generic time-interval selection. As motion logs may be noisy we will likely also need to think about thresholds, etc., or a way to define, e.g., a static position from a noisy position log (raising an error if it cannot be considered static).

PR #114, opened 2024-10-02
Expose NeXus transformation chains in workflow steps

SimonHeybrock:
Part of #96.

This is not completing this task, but does most of the prep work. The idea is that the `to_transformation` could accept additional arguments (or be replaced by one that does), which could:
1. Select a specific time interval, including averaging and threshold checking.
2. Update values of one or more transformations in the chain. This is mainly relevant for live-data reduction.

Neither 1.) nor 2.) is implemented right now. The initial focus will be on 1.). I would mainly like input from the reviewer on whether this new structure is adequate for this. The actual implementation of these features could be performed in follow-up work.

Depends on a release of scipp/scippnexus#243 (that is why CI is failing for now, otherwise this PR is ready for review).

Note that I ran into major (code) overhead/duplication from the current duplication (or near-duplication) of different code for different component types. Previously we had failed to find a way to generalize this. The reason was that monitors required one extra type-parameter (`MyMonitor[RunType, MonitorType]` vs `MyDetector[RunType]`, `MySample[RunType]`, ...). I believe I have now solved this via the new `ComponentType`. We can thus use `MyComponent[Component, RunType]` almost throughout, resulting in a significant deduplication. The "trick" is that `Component` takes general component names except for the monitors, where there are multiple named ones.


src/ess/reduce/nexus/_nexus_loader.py
```diff
@@ -41,10 +41,15 @@ def load_component(
         component = _unique_child_group(instrument, nx_class, group_name)
         loaded = cast(sc.DataGroup, component[selection])
         loaded['nexus_component_name'] = component.name.split('/')[-1]
-    return compute_component_position(loaded)
+    return loaded
```
SimonHeybrock, line 44: This is the heart of the change in this PR: We are no longer computing the position directly here. This is moved into an extra step.

src/ess/reduce/nexus/types.py
```diff
@@ -95,9 +91,9 @@ class TransmissionRun(Generic[ScatteringRunType]):
 """Identifier for an arbitrary monitor"""
 Monitor5 = NewType('Monitor5', int)
 """Identifier for an arbitrary monitor"""
-Incident = NewType('Incident', int)
+IncidentMonitor = NewType('IncidentMonitor', int)
```
SimonHeybrock, line 94: Added the monitor suffix here since these now show up as `MyComponent[IndicidentMonitor]` and it may be confusing without.

src/ess/reduce/nexus/workflow.py
```diff
@@ -204,102 +147,62 @@ def load_nexus_sample(
...
-        nexus.load_component(location, nx_class=snx.NXdetector, definitions=definitions)
-    )
-
-
-def load_nexus_monitor(
-    location: NeXusMonitorLocationSpec[RunType, MonitorType],
-) -> NeXusMonitor[RunType, MonitorType]:
-    """
-    Load monitor from NeXus, but with event data replaced by placeholders.
+    When loading a detector or monitor, event data is replaced by placeholders.
```
nvaytet, line 177: Maybe I missed something, but is this for loading only detectors and monitors? I thought `Component` was also for sample and source?

**Edit**: I now see that the case for `NXsample` is overloaded by the `load_nexus_sample` function above. Are we missing a `load_nexus_source` function?

src/ess/reduce/nexus/workflow.py
```diff
@@ -552,34 +468,35 @@ def _add_variances(da: sc.DataArray) -> sc.DataArray:
 definitions["NXmonitor"] = _StrippedMonitor
 
 
-_common_providers = (gravity_vector_neg_y, file_path_to_file_spec, all_pulses)
+_common_providers = (
+    gravity_vector_neg_y,
+    file_path_to_file_spec,
+    all_pulses,
+    component_spec_by_name,
+    unique_component_spec,  # after component_spec_by_name, partially overrides
```
nvaytet, line 476: Does the comment mean that we are relying on the behaviour of the pipeline; that it applies the providers in the order that they are given. Can we assume that the behaviour will never change?

src/ess/reduce/nexus/workflow.py
SimonHeybrock: It is also used for the source.  But this paragraph basically just applies to the groups that have event data.

src/ess/reduce/nexus/workflow.py
SimonHeybrock: There are explicit tests for this, e.g., here: https://github.com/scipp/sciline/blob/a90b73c50d4cef1fc8adb231dc2208cf99ef2f0a/tests/pipeline_test.py#L916. That is, this is intentional behavior and not likely to change. If it would, quite a few other workflows would also break, I think.

src/ess/reduce/nexus/workflow.py
```diff
@@ -204,102 +147,62 @@ def load_nexus_sample(
...
+        dg = sc.DataGroup(depends_on=snx.TransformationChain(parent='', value='.'))
+    return NeXusComponent[snx.NXsample, RunType](dg)
 
 
-def load_nexus_source(
-    location: NeXusComponentLocationSpec[snx.NXsource, RunType],
-) -> NeXusSource[RunType]:
-    """
-    Load a NeXus source group from a file.
+def nx_class_for_monitor() -> NeXusClass[MonitorType]:
```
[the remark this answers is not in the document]
nvaytet: I think it's fine as is for now.

nvaytet (approved):

PR #115, opened 2024-10-04
Add "JSON" generator to facilitate testing of live workflows in downstream packages

SimonHeybrock:
For now this only supports NXevent_data. More might be added in the future.

### DIFF
--- a/src/ess/reduce/nexus/json_generator.py
+++ b/src/ess/reduce/nexus/json_generator.py
@@ -8,7 +8,7 @@ from typing import Any
 import scipp as sc
 
 
-def _as_attr(value: Any, name: str) -> dict:
+def _as_str_attr(value: Any, name: str) -> dict:
     val = str(value)
     return {"string_size": len(val), "type": "string", "name": name, "values": val}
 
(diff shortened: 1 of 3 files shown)


tests/nexus/json_generator_test.py
```diff
@@ -0,0 +1,38 @@
...
+from ess.reduce.nexus.json_nexus import json_nexus_group
+
+
+def test_event_data_generator_monitor_events_round_trip() -> None:
+    filename = data.loki_tutorial_sample_run_60250()
+    monitor = snx.load(filename, root='entry/instrument/monitor_1/monitor_1_events')
+    generator = event_data_generator(monitor)
+    for i in range(len(monitor)):
+        group = json_nexus_group(next(generator))
+        assert_identical(group[()], monitor[i : i + 1])
```
jl-wynen, line 17: Can you add a test that the generator has finished after the loop? I.e., that `next(generator)` raises `StopIteration`. Asking because the loop uses `monitor` and not `generator`.

src/ess/reduce/nexus/json_generator.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def event_data_generator(data: sc.DataArray) -> Generator[dict, None, None]:
+    """
+    Generate JSON data for event data from a NeXus file.
+
+    Parameters
+    ----------
+    data:
+        A data array with event data, equivalent to what ScippNexus would load from an
```
jl-wynen, line 69: It's a dict, not a data array.

SimonHeybrock: Hmm, do you mean the return value? This docstring line is about the input.

jl-wynen: Yes, I meant the return value. My bad.
Can you add an explanation of it? In particular, there is no explanation of how data is encoded.

jl-wynen (approved):

PR #118, opened 2024-10-10
Add `ess.reduce.live`, a module for building live reduction workflows

SimonHeybrock:
This is a very early version, and I expect several cycles of additions and modifications. I am using this with so far 3 workflows in ESSsans, which has driven/informed what should go into this module. The `live` module uses the `streaming` module, but operates on a slightly higher level and is more tailored towards Beamlime.

### DIFF
--- a/src/ess/reduce/live.py
+++ b/src/ess/reduce/live.py
@@ -38,6 +38,8 @@ T = TypeVar('T', bound='LiveWorkflow')
 
 
 class LiveWorkflow:
+    """A workflow class that fulfills Beamlime's LiveWorkflow protocol."""
+
     def __init__(
         self,
         *,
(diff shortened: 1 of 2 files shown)


src/ess/reduce/live.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Returns
+        -------
+        :
+            Dictionary of computed and plottable results.
+        """
+        # Beamlime passes full path, but the workflow only needs the name of the monitor
+        # or detector group.
+        nxevent_data = {
+            key.lstrip('/').split('/')[2]: value for key, value in nxevent_data.items()
```
nvaytet, line 123: Is there a less brittle way we can make a key without relying on the number `2`?

nvaytet (commented):
Looks good, just one small comment.

src/ess/reduce/live.py
SimonHeybrock: I think all NeXus files I have seen have monitors and detectors in `/entry/instrument/<name>`. Can't think of a simple option that will not be even more brittle. Let us fix it when it causes trouble?

nvaytet (approved):

PR #119, opened 2024-10-10
Adapt NeXus workflow to default to sample at the origin if sample group has no position

SimonHeybrock:
Previously we created a default sample at the origin if not found in the file. With this change, we handle the case where a sample group is present but without position. Defaulting to origin there as well.

### DIFF
--- a/src/ess/reduce/nexus/workflow.py
+++ b/src/ess/reduce/nexus/workflow.py
@@ -147,7 +147,9 @@ def load_nexus_sample(
     try:
         dg = nexus.load_component(location, nx_class=snx.NXsample)
     except ValueError:
-        dg = sc.DataGroup(depends_on=snx.TransformationChain(parent='', value='.'))
+        dg = sc.DataGroup()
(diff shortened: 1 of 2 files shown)


src/ess/reduce/nexus/workflow.py
```diff
@@ -147,7 +147,9 @@ def load_nexus_sample(
     try:
         dg = nexus.load_component(location, nx_class=snx.NXsample)
     except ValueError:
-        dg = sc.DataGroup(depends_on=snx.TransformationChain(parent='', value='.'))
+        dg = sc.DataGroup()
+    if 'depends_on' not in dg:
+        dg['depends_on'] = snx.TransformationChain(parent='', value='.')
```
nvaytet, line 152: Should we add a test for this use case?

SimonHeybrock: See update.

nvaytet (approved):

PR #132, opened 2024-11-18
Copier update

jl-wynen:


.copier-answers.yml
```diff
@@ -1,5 +1,5 @@
 # Changes here will be overwritten by Copier; NEVER EDIT MANUALLY
-_commit: bd9d844
+_commit: aa5dc5e
 _src_path: gh:scipp/copier_template
 description: Common data reduction tools for the ESS facility
 max_python: '3.12'
```
MridulS, line 5: Should we bump to python 3.13? any blockers?

jl-wynen: Technically, not all dependencies are tagged for 3.13 yet. But that does not actually matter for installing anything. So I updated.

MridulS (approved):

PR #133, opened 2024-11-18
Load choppers and analysers

jl-wynen:
This extends the component-loading code to allow loading all components of a certain type. With choppers and analysers, we typically need all of them. So there is little use in splitting it up and loading each one individually. Also, BIFROST has a lot of analysers and I would rather not list all their names in a call to map-reduce.

Note that choppers are not parsed into `scippneutron.chopper.DiskChopper` objects because their time-dependent quantities first need to be processed. And that can probably not be done generically in ESSreduce.


src/ess/reduce/nexus/_nexus_loader.py
```diff
@@ -28,22 +28,41 @@ def load_component(
...
+    *,
+    nx_class: type[snx.NXobject],
+    definitions: Mapping | None | NoNewDefinitionsType = NoNewDefinitions,
+) -> sc.DataGroup:
+    """Load all components of a given class from NeXus."""
+    if location.component_name is not None:
+        raise ValueError(
+            "`location.component_name` must be None when loading all "
+            "components of a certain class."
+        )
```
SimonHeybrock, line 54: Why are we using that dataclass here in the first place?

src/ess/reduce/nexus/workflow.py
```diff
@@ -96,6 +100,22 @@ def component_spec_by_name(
...
+    return NeXusComponentLocationSpec[snx.NXdisk_chopper, RunType](
+        filename=filename.value
+    )
+
+
+def crystal_component_spec(
+    filename: NeXusFileSpec[RunType],
+) -> NeXusComponentLocationSpec[snx.NXcrystal, RunType]:
+    """Create a location spec for a crystal / analyzer group in a NeXus file."""
+    return NeXusComponentLocationSpec[snx.NXcrystal, RunType](filename=filename.value)
```
SimonHeybrock, line 116: Feels redundant, since there is no component name, so what is the point in distinguishing between component types? Should we just have `NeXusEntrySpec[RunType]` or something like that, with not component name?

src/ess/reduce/nexus/workflow.py
```diff
@@ -465,6 +514,36 @@ def assemble_monitor_data(
...
+    """Convert the NeXus representation of a chopper to ours."""
+    return Choppers[RunType](
+        sc.DataGroup(
+            {
+                name: extract_chopper_from_nexus(
+                    nexus.compute_component_position(chopper)
+                )
+                for name, chopper in choppers.items()
+            }
+        )
```
SimonHeybrock, line 529: Can this use https://scipp.github.io/generated/classes/scipp.DataGroup.html#scipp.DataGroup.apply? Feel free to ignore this comment unless shorter.

SimonHeybrock (commented):
Looks good overall. Minor questions/comments (most can be ignored):

src/ess/reduce/nexus/workflow.py
jl-wynen: Yes. But it only saves 1 line here. It's better for the analysers, though.

SimonHeybrock (approved):

PR #146, opened 2024-12-10
fix: support non-html outputs

### LINKED ISSUE #144: GUI: Support output of any type
Currently, the workflow widget output requires that all outputs have `_html_repr_`. This is not always the case: https://github.com/scipp/essreduce/issues/143. We should fully use IPython's output system to use whatever representation works best.

jokasimr:
Fixes #144

### DIFF
--- a/src/ess/reduce/ui.py
+++ b/src/ess/reduce/ui.py
@@ -130,7 +130,7 @@ class ResultBox(widgets.VBox):
                     result_registry.clear()
                     result_registry.update(compute_result)
                 for i in compute_result.values():
-                    display.display(display.HTML(i._repr_html_()))
+                    display.display(i)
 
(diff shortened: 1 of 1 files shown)


jl-wynen:
Does this work with all Python objects? Or does IPython have requirements for the arguments of `display`?

Is there a way to test this (and the UI code in general) automatically? At least to ensure it doesn't raise exceptions.

jokasimr:
The way I understand it IPython has a preferred order of `_repr_`s. If the object and the environment supports a more sophisticated repr then that will be used.

YooSunYoung (approved):
I think it's good to go....?

jl-wynen:
This fixes something that doesn't already have a test, so I'd like to see a new test.

jl-wynen (approved):

PR #147, opened 2024-12-10
feat: add simple progress indicator

### LINKED ISSUE #143: GUI: Progress indicator
We can in principle show a progress bar or something like it for a running workflow by showing how many providers have completed.

If this is too difficult / undesired, we could at least at something to indicate whether the widget is busy running the pipeline right now.

jokasimr:
Fixes #143

### DIFF
--- a/src/ess/reduce/widgets/_spinner.py
+++ b/src/ess/reduce/widgets/_spinner.py
@@ -1,0 +1,100 @@
+_spinner = '''
+<style>
+.lds-spinner {
+  /* change color here */
+  color: #1c4c5b
+}
+.lds-spinner,
+.lds-spinner div,
+.lds-spinner div:after {
+  box-sizing: border-box;
+}
+.lds-spinner {
+  color: currentColor;
+  display: inline-block;
+  position: relative;
+  width: 80px;
+  height: 80px;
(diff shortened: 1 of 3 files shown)


YooSunYoung:
@jokasimr docs build often hangs and it looked like hanging again so I canceled/rerun the job.
And it seems like there is a merge conflict...?

jokasimr:
@YooSunYoung the merge conflict is fixed now.
Do you think the hanging is caused by this change or is that something old?

src/ess/reduce/widgets/_spinner.py
```diff
@@ -0,0 +1,100 @@
+_spinner = '''
+<style>
```
nvaytet, line 2: Would using a pre-made spinner be a thing we could use?
e.g. https://fontawesome.com/v4/icon/spinner

jokasimr: We could, there are tons of options out there. The one currently implemented was just copied from a long list of spinner alternatives.

Using font-awesome seems to require downloading their font package and that seems overkill for this little functionality.

nvaytet (approved):

Issue #74, opened 2024-08-13
Improve Switchable Widget.


YooSunYoung:

jokasimr:
I was thinking the "select parameters" section could just be one big multiple-select containing both parameters and intermediate parameters all together (all domain types in the workflow).

Then when a workflow and a set of outputs is selected the root parameters of the graph would be selected. If the user wants to remove some of them and set some intermediary values manually they make that change in the parameters multiple-select and the parameters form gets updated.

SimonHeybrock:
One problem with this solution is that it is totally non obvious to the user which parameters can be set. Parts of defining "THE" parameter of the workflow is to document which inputs are required or options. Saying "any of the intermediate results of the workflow is an optional input" does not serve this important purpose.

SimonHeybrock:
Then I do not understand the suggestion. For example, above it says

> Instead of setting switchable to all the parameters, we can treat all parameters switchable, and have a menu of parameter from the workflow, just like we select output.

Can you give an example of how this would work for, say, https://github.com/scipp/esssans/blob/1c407162a50f4e68b843899f55384f5c025e19e7/src/ess/sans/parameters.py#L99-L104?

SimonHeybrock:
Ok, so we keep the `Parameter.switchable` field 👍

jokasimr:
> Ok, so we keep the Parameter.switchable field 👍

I think @YooSunYoung point was that removing switchable widget would make it easier to style the parameter form.

SimonHeybrock:
I was refering to `class Parameter`, not the widget itself. Are you suggesting that there should be a different mechanism outside `Parameter` that defines which parameters or the workflow are switchable?

jokasimr:
> I was refering to class Parameter, not the widget itself. Are you suggesting that there should be a different mechanism outside Parameter that defines which parameters of the workflow are switchable by default?

No I don't have an opinion on that

jokasimr:
For clarity here are a couple of screenshots illustrating the idea:

![Screenshot from 2024-08-13 16-08-50](https://github.com/user-attachments/assets/86d58c12-8f0e-48d0-acdc-ba64b00b625f)
![Screenshot from 2024-08-13 16-09-09](https://github.com/user-attachments/assets/e3f5e636-ecb8-4f4d-8981-7c06e41abbe6)

SimonHeybrock:
[...]

PR #148, opened 2024-12-10
Switch/Optional widget fields setter/getter

### LINKED ISSUE #140: Saving parameters and load them
We want to let users save parameters and auto-complete the parameter inputs from it.

YooSunYoung:
Fixes #140 

I just merged all smaller PRs here after all since they all need each other to understand...

Here are review points

# Treat `value` as one of fields.

The dictionary that is exported from the `ParameterBox` look like:
```python
{
  type_object: { field_name_1: field_value_1 , field_name_2: field_value_2 }
}
```

At first I thought if the input widget doesn't have a field but a value, it should be like
```python
{
  type_object: input_widget_value
}
```

But it made it difficult for wrapping widget like `OptionalWidget` or `SwitchWidget` to add their states into the exported dictionary so I updated the exported dictionary to be either

```python
{
  type_object_1: { field_name_1: field_value_1 , field_name_2: field_value_2 },
  type_object_2: { 'value': input_widget_value }
}
```

So that wrapper widget can extend the dictionary with their state, 
i.e. `SwitchWidget` has `enabled` so it'll look like
```python
{
  type_object_1: { 'enabled': True/False, field_name_1: field_value_1 , field_name_2: field_value_2 },
  type_object_2: { 'enabled': True/False, 'value': input_widget_value }
}
```

It picks up the wrapper widget state key and forward the rest to the `wrapped` widget.

# SwitchWidget

As explained above, `SwitchWidget` extends the widget state with `enabled` field.

# OptionalWidget

Similar to the `SwitchWidget`, it extends the widget state with `opted-out` field.

The reason why it's not setting `value` as `None` instead of using `opted-out` flag is explained in the comment in the code.

# `enabled` and `opted-out` fields.

The `set_fields/get_fields` methods of `SwitchWidget` and `OptionalWidget` are implemented assuming
the `wrapped` widget don't use the `enabled` or `opted-out` for setting/getting their fields.

I thought it's safe to do so because `SwitchWidget` will always control the `availability` of the `wrapped` widget, 
so any other widgets will need to export/set `enabled/disabled` state themselves.


tests/widget_test.py
```diff
@@ -423,3 +467,73 @@ def run_workflow():
...
+    widget.enabled = True
+    assert widget.get_fields() == {'enabled': True, 'x': 1, 'y': 2, 'z': 3, 'unit': 'm'}
+    widget.set_fields({'enabled': False, 'x': 4, 'y': 5, 'unit': 'mm'})
+    assert widget.get_fields() == {
+        'enabled': False,
+        'x': 4,
+        'y': 5,
+        'z': 3,
+        'unit': 'mm',
+    }
```
nvaytet, line 539: Probably a stupid question but do we really need the switchable and optional widgets?
I found their usability confusing, and their behaviour sometimes inpredictable (e.g. in the Loki workflow, when toggling a switchable widget on (direct beam filename), sometimes it is then set to a default value, and sometimes it has no value set)

Can we do without? It would simplify the code.

YooSunYoung: So the `optional` widget is for setting `None`
and `switchable` is for setting it as a parameter, which effectively prune a graph from the workflow.

The usability indeed needs improvement, so we have a issue about that https://github.com/scipp/essreduce/issues/74

But I don't think we can't simply remove them to support those use cases I mentioned at the beginning.

If you want to remove them from the specific workflow, you can just set ``optional`` or ``switchable`` from the parameter.

src/ess/reduce/widgets/_optional_widget.py
```diff
@@ -64,3 +65,27 @@ def value(self, value: Any) -> None:
         else:
             self._option_box.value = self.name
             self.wrapped.value = value
+
+    def set_fields(self, new_values: dict[str, Any]) -> None:
+        new_values = dict(new_values)
+        # Set the value of the option box
+        opted_out_flag = (
+            new_values.pop(  # We assume ``opted-out`` is not used in any wrapped widget
+                'opted-out',
```
nvaytet, line 74: Is `opted-out` a safe enough name? I seems rather safe, but should we make it even less likely to clash? e.g. `essreduce-opted-out` or something like that? (in the past with plopp we often added the name of the library in such attributes).

I'm guessing the `opted-out` is never seen by users, so the cleanliness of the name doesn't really matter?

src/ess/reduce/widgets/_switchable_widget.py
```diff
@@ -49,3 +50,17 @@ def value(self) -> Any:
...
+
+    def set_fields(self, new_values: dict[str, Any]) -> None:
+        # Retrieve and set the enabled flag first
+        new_values = dict(new_values)
+        enabled_flag = new_values.pop('enabled', self.enabled)
+        if not isinstance(enabled_flag, bool):
+            raise ValueError(f"`enabled` must be a boolean, got {enabled_flag}")
+        self.enabled = enabled_flag
+        # Set the rest of the fields
+        set_fields(self.wrapped, new_values)
```
nvaytet, line 62: Should we only fill the rest of the fields if the widget is enabled, or do we fill in anyway?

tests/widget_test.py
nvaytet: I think I still don't really understand the difference between optional and switchable. They seem very similar: if enabled, use the value?
Can't we manage with only one of the two?
(sorry for being slow here)

src/ess/reduce/widgets/_switchable_widget.py
YooSunYoung: We should restore the state even if they are disabled. So that user can easily switch it on and see the restored status.

tests/widget_test.py
YooSunYoung: @nvaytet  No worries, we should have had a better documentation from the first place... 
I'll first write down some description here and update the documentation.

# `Switchable` widget
is about setting values as parameters of the workflow (`sciline.Pipeline` object )
  - if `enabled`, it set the value as the parameter before it computes things
     it could mean it set an intermediate result and prune the graph.
  - if `disabled` it is ignored and not seen by the workflow object.

# `Optional` widget  
is about setting `None` or a meaningful value.
Some of providers are setting a local variable if an argument is `None` for example... https://github.com/scipp/esssans/blob/05b6521213150ff89caa75dac00406436607df02/src/ess/sans/normalization.py#L268-L271

So in any cases `Optional` widget value will be set as a parameter when `compute` is requested.
We needed a radio button or sth similar to it because the actual meaningful value and the `None` can't be represented by a same widget.

nvaytet (approved):
Sorry for being extremely slow to come back to this (it slipped under my radar).

I think we should just move ahead with this as I don't really have issues with the PR itself, it was more conceptual questions on a higher level.

nvaytet:
@YooSunYoung should we get this merged?

PR #155, opened 2024-12-13
Extract function to prune type vars in workflow

jl-wynen:
When building a concrete workflow, we usually modify the result of `GenericNeXusWorkflow()`. This reintroduces run types and monitor types that are excluded by `run_types` and `monitor_types`.

Extracting `prune_nexus_domain_types` into its own function allows us to remove unused types after inserting custom providers into the workflow.

### DIFF
--- a/src/ess/reduce/nexus/workflow.py
+++ b/src/ess/reduce/nexus/workflow.py
@@ -3,11 +3,9 @@
 
 """Workflow and workflow components for interacting with NeXus files."""
 
-from collections.abc import Sequence
 from copy import deepcopy
 from typing import Any
 
-import networkx as nx
 import sciline
 import sciline.typing
 import scipp as sc
@@ -659,7 +657,7 @@ def GenericNeXusWorkflow() -> sciline.Pipeline:
 
     See Also
     --------
-    prune_nexus_domain_types:
+    ess.reduce.workflow.prune_nexus_domain_types:
         Use this function after creating and customizing a
         NeXus workflow to avoid very large graphs.
     """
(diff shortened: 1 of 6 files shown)


src/ess/reduce/nexus/workflow.py
```diff
@@ -647,32 +648,21 @@ def LoadDetectorWorkflow() -> sciline.Pipeline:
     return wf
 
 
-def GenericNeXusWorkflow(
-    *,
-    run_types: Sequence[sciline.typing.Key] | None = None,
-    monitor_types: Sequence[sciline.typing.Key] | None = None,
-) -> sciline.Pipeline:
+def GenericNeXusWorkflow() -> sciline.Pipeline:
```
SimonHeybrock, line 651: This change will break downstream such as ESSsans which is using these args already. Is this deliberate?

jl-wynen: Deliberate in that it is easier than keeping the arguments in two places.

But also, then it breaks, the error should indicate to us that the code should be changed. As it stand, it suffers from the exact problem that I am trying to fix.

Note that I converted this to a draft because it does not actually work with a full workflow. I am adding a way to specify target types instead of hard coding `DetectorData` and `MonitorData`.

jl-wynen:
This is ready now. It works in ESSdiffraction: https://github.com/scipp/essdiffraction/pull/115 It should also work in ESSspectroscopy where we have multiple custom parametrized types.

There is a distinct possibility that mypy is unhappy with how I used generic types in `_add_ancestors`. But if we ever get this far with type checking, we can just silence mypy in this function :)

src/ess/reduce/workflow.py
```diff
@@ -91,3 +93,139 @@ def assign_parameter_values(pipeline: Pipeline, values: dict[Key, Any]) -> Pipel
...
+    targets_per_run:
+        Types parametrized by run type to keep.
+    targets_per_run_and_monitor:
+        Types parametrized by run type and monitor type to keep.
+    run_types:
+        List of run types to include in the workflow. If not provided, all run types
+        are included.
+    monitor_types:
+        List of monitor types to include in the workflow. If not provided, all monitor
+        types are included.
```
SimonHeybrock, line 131: Before looking at this more closely, can we have a chat about the new interface? I feel the need to specify targets means that this function (and pruning in general) will be unusable by anyone not intricately familiar with a workflow. 

Furthermore, doesn't it mean that every time a workflow is changed there is a risk that all the pruning calls break (and need to be updated)? It may even become hard to see through, e.g., if someone adds a new target (towards the end of a workflow) but the workflow it refuses to compute it because it was unintentionally pruned.

SimonHeybrock:
As far as I understand this has been superseded.

PR #157, opened 2024-12-16
Prune type vars v2

jl-wynen:
Supersedes #155

### DIFF
--- a/src/ess/reduce/nexus/workflow.py
+++ b/src/ess/reduce/nexus/workflow.py
@@ -3,12 +3,12 @@
 
 """Workflow and workflow components for interacting with NeXus files."""
 
-from collections.abc import Sequence
+from collections.abc import Iterable
 from copy import deepcopy
 from typing import Any
 
-import networkx as nx
 import sciline
+import sciline.typing
 import scipp as sc
 import scippnexus as snx
 from scipp.constants import g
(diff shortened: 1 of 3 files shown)


src/ess/reduce/nexus/workflow.py
```diff
@@ -685,16 +688,34 @@ def GenericNeXusWorkflow(
...
+) -> None:
+    # Remove all nodes that use a run type or monitor types that is
+    # not listed in the function arguments.
+    excluded_run_types = _excluded_type_args(RunType, run_types)
+    excluded_monitor_types = _excluded_type_args(MonitorType, monitor_types)
+    excluded_types = excluded_run_types | excluded_monitor_types
+
+    graph = workflow.underlying_graph
+    to_remove = [
+        node for node in graph if excluded_types & set(getattr(node, "__args__", set()))
```
SimonHeybrock, line 711: Does this work with nesting? One example to consider is the ESSsans `TransmissionRun`, e.g., in `NeXusComponent[IncidentMonitor, TransmissionRun[SampleRun]]`. But there can also be other types of nesting, I think?

jl-wynen: Works in SANS: https://github.com/scipp/esssans/pull/190

SimonHeybrock: So we will assume that no other nesting is needed? That is, we will not support, say `dict[str, Filename[RunType]]`?

SimonHeybrock: In this concrete example it is ok. And I am not aware of others.

We can also postpone nesting support until later, if required?

jl-wynen: Yes, I would say so

SimonHeybrock (approved):

PR #163, opened 2025-01-16
Use Monte-Carlo simulations for frame unwrapping and WFM

nvaytet:
## Moving frame unwrapping and WFM processing from `scippneutron` to `essreduce`

In this PR, we move away from our analytical computations of frame boundaries and fitting the polygons with single lines for frame unwrapping and WFM.

This method yielded errors typically towards the edge of the frames where the polygon fit would be less accurate (the lines would in some cases be outside of the polygons). Using a 3rd order fit helped but still wasn't optimal.

In this work, we instead use the `tof` package to illuminate the chopper cascade with a pulse of neutrons.
We then propagate the neutrons that exit the cascade to the distances of the pixels (this is just a single broadcast operation) and build a lookup table that gives us wavelength (or time-of-flight) as a function of distance and time-of-arrival.

The results are better than with the previous method (tested on a DREAM simulation), and the code simpler.
The computation is somewhat more expensive than before, but for now it is not very noticeable.

More importantly, this opens the door to other simulation methods. Instead of `tof`, one could in the future use McStas to make a much more realistic simulation of neutrons travelling through the cascade, including effects like non-instantaneous opening of choppers or neutron reflections along the beam guide.

See the `frame-unwrapping` and `wfm-time-of-flight` notebooks for a description of how the algorithm works.

**Comparison with previous version:**

Plots of dspacing vs 2theta:

- `true`: using the true wavelengths of the neutrons from the simulation data
- `from_tof`: the current version implemented in this PR
- `from_main`: the version on scippneutron/main using linear fit to polygons
- `from_cubic`: an (unpublished) improvement on scippneutron/main using a cubic fit to polygons (see https://github.com/scipp/scippneutron/tree/cubic-polygons)

Screenshot_20250114_141344

1D dspacing plot:

Figure 1 (16)
Figure 1 (17)


src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,451 @@
...
+    Extract the total length of the flight path from the source to the detector from the
+    detector data.
+
+    Parameters
+    ----------
+    da:
+        Raw detector data loaded from a NeXus file, e.g., NXdetector containing
+        NXevent_data.
+    """
+    return Ltotal[RunType](da.coords["Ltotal"])
```
jl-wynen, line 85: Similarly to the source, does it make sense to have a Beamline class / module to encode common data? I am also thinking of metadata like the beamline's name, facility, etc, which would be useful for writing output files. See also https://github.com/scipp/scippneutron/issues/473

jl-wynen (commented):
I only looked at high level structure as I am unfamiliar with the algorithmic details.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,451 @@
...
+
+    # Need to add the simulation distance to the distance coordinate
+    wavelength.coords["distance"] = wavelength.coords["distance"] + simulation_distance
+    h = sc.constants.h
+    m_n = sc.constants.m_n
+    velocity = (h / (wavelength * m_n)).to(unit="m/s")
+    timeofflight = (sc.midpoints(wavelength.coords["distance"])) / velocity
+    out = timeofflight.to(unit=time_unit, copy=False)
+    # Include the variances computed above
+    out.variances = variance.values
```
jokasimr, line 154: Aren't the variances the variance of the wavelength? 
It seems strange to set the time-of-flight variance equal to the wavelength variance.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,451 @@
...
+    ltotal:
+        Total length of the flight path from the source to the detector.
+    toas:
+        Time of arrival of the neutron at the detector, folded by the frame period.
+    """
+    from scipy.interpolate import RegularGridInterpolator
+
+    lookup_values = lookup.data.to(unit=elem_unit(toas), copy=False).values
+    # Merge all masks into a single mask
+    if lookup.masks:
```
jokasimr, line 340: Would it make sense to do this in the provider that creates `MaskedTimeOfFlightLookup`? It seems odd to modify the lookup table in a function that uses it rather than produces it.

src/ess/reduce/time_of_flight/fakes.py
```diff
@@ -0,0 +1,241 @@
...
+        beam_position=ch.beam_position,
+        phase=ch.phase,
+        axle_position=ch.axle_position,
+        slit_begin=ch.slit_begin[0:1],
+        slit_end=ch.slit_end[0:1],
+        slit_height=ch.slit_height[0:1],
+        radius=ch.radius,
+    )
+    for name, ch in wfm_choppers.items()
+}
```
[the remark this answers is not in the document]
nvaytet: We could move them, but I would say those choppers are meant to be used with the `FakeBeamline`, so I would leave them there?

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,451 @@
...
+    Return the period of the source pulses, i.e., time between consecutive pulse starts.
+
+    Parameters
+    ----------
+    facility:
+        Facility where the experiment is performed (used to determine the source pulse
+        parameters).
+    """
+    facilities = {"ess": sc.scalar(14.0, unit="Hz")}
+    return PulsePeriod(1.0 / facilities[facility])
```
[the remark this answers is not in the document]
nvaytet: Keeping a database of useful data would be useful, also the wavelength range of the source, or things like a rough range of distances covered by Ltotal for each instrument?
But maybe we can delay that to another PR?

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,451 @@
...
+            "toa": toas.flatten(to="event"),
+            "wavelength": sc.broadcast(simulation.wavelength, sizes=toas.sizes).flatten(
+                to="event"
+            ),
+            "distance": sc.broadcast(distances, sizes=toas.sizes).flatten(to="event"),
+        },
+    )
+
+    # TODO: move toa resolution to workflow parameter
+    binned = data.bin(distance=ndist, toa=500)
```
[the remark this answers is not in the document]
nvaytet: Thanks for pointing this out. I refactored the distances
There is a bit of funny stuff going on with bin edges and midpoints, when binning data and then sending to the bilinear interpolator, but I think I got it right now.

src/ess/reduce/time_of_flight/toa_to_tof.py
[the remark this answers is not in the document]
jl-wynen: > But maybe we can delay that to another PR?

Definitely. Just wanted to let you know about it.

src/ess/reduce/time_of_flight/fakes.py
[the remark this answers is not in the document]
jl-wynen: If they are useful outside of ESSreduce / ScippNeutron, yes. But can you actually use them in any other package?

src/ess/reduce/time_of_flight/toa_to_tof.py
nvaytet: Ok. I figured out what my mistake was. I do get the same for both. I will use that instead.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,520 @@
...
+    """
+    return Ltotal(da.coords["Ltotal"])
+
+
+def compute_tof_lookup_table(
+    simulation: SimulationResults,
+    ltotal_range: LtotalRange,
+    distance_resolution: DistanceResolution,
+    toa_resolution: TimeOfArrivalResolution,
+) -> TimeOfFlightLookupTable:
```
SimonHeybrock, line 99: docstring

src/ess/reduce/time_of_flight/types.py
```diff
@@ -0,0 +1,154 @@
...
+class SimulationResults:
+    """
+    Results of a time-of-flight simulation used to create a lookup table.
+    """
+
+    time_of_arrival: sc.Variable
+    speed: sc.Variable
+    wavelength: sc.Variable
+    weight: sc.Variable
+    distance: sc.Variable
```
SimonHeybrock, line 48: Would be useful to have some docs about the fields here. The generic `Variable` type hint does not say much about, e.g., shape and dims.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,520 @@
...
+
+def find_fastest_neutron(simulation: SimulationResults) -> FastestNeutron:
+    """
+    Find the fastest neutron in the simulation results.
+    """
+    ind = np.argmax(simulation.speed.values)
+    return FastestNeutron(
+        time_of_arrival=simulation.time_of_arrival[ind],
+        speed=simulation.speed[ind],
+        distance=simulation.distance,
```
SimonHeybrock, line 201: No index? The `SimulationResults` docstring is not very telling.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,520 @@
...
+    da: RawData,
+    lookup: MaskedTimeOfFlightLookupTable,
+    ltotal: Ltotal,
+    toas: FrameFoldedTimeOfArrival,
+) -> TofData:
+    from scipy.interpolate import RegularGridInterpolator
+
+    # TODO: to make use of multi-threading, we could write our own interpolator.
+    # This should be simple enough as we are making the bins linspace, so computing
+    # bin indices is fast.
```
SimonHeybrock, line 347: Implying that this function is the slow part of the entire process?

Can you give some benchmark results? How many events/seconds can this process? This is important to know as it will impact everything, so we can plan for extra effort here, if required.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,520 @@
...
+    Default parameters of the time-of-flight workflow.
+    """
+    return {
+        PulseStride: 1,
+        PulseStrideOffset: 0,
+        DistanceResolution: sc.scalar(0.1, unit="m"),
+        TimeOfArrivalResolution: 500,
+        LookupTableRelativeErrorThreshold: 1.0e-2,
+        SimulationSeed: 1234,
+        NumberOfNeutrons: 1_000_000,
```
SimonHeybrock, line 424: I am guessing, but it looks like some of the parameters are specific to the `tof` simulation? Can we split this up?

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,520 @@
...
+        pulse_stride_offset=None,
+        distance_resolution=None,
+        toa_resolution=None,
+        error_threshold=None,
+        seed=None,
+        number_of_neutrons=None,
+    ):
+        import sciline as sl
+
+        self.pipeline = sl.Pipeline(standard_providers())
```
SimonHeybrock, line 479: I suggest changing this class so the simulation providers are dependency-injected in the constructor. This will allow for easily swapping it out for a different simulation.

src/ess/reduce/time_of_flight/types.py
```diff
@@ -0,0 +1,154 @@
...
+
+LtotalRange = NewType("LtotalRange", tuple[sc.Variable, sc.Variable])
+"""
+Range (min, max) of the total length of the flight path from the source to the detector.
+"""
+
+
+DistanceResolution = NewType("DistanceResolution", sc.Variable)
+"""
+Resolution of the distance axis in the lookup table.
```
SimonHeybrock, line 70: Unclear what exactly "resolution" is here. A length giving the bin size?

src/ess/reduce/time_of_flight/fakes.py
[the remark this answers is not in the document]
nvaytet: I guess something that can create fake beamline data that looks like semi-realistic NeXus data could be used by other packages for testing? (even though I don't have anything specific in mind right now)

src/ess/reduce/__init__.py
```diff
@@ -4,7 +4,7 @@
 
 import importlib.metadata
 
-from . import nexus, uncertainty
+from . import nexus, uncertainty, time_of_flight
```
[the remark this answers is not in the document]
nvaytet: Looks very good. I wasn't aware of the `lazy_loader`.
I am guessing you meant for a later stage? i.e. not part of this PR?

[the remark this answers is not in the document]
SimonHeybrock: Yes I suppose. Unless needed you can also consider not importing the submodule here, to avoid worrying about making some dependencies mandatory for downstream packages.

src/ess/reduce/time_of_flight/to_events.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Create the data counts, which are the original counts divided by the number of
+    # events per bin
+    sizes = {event_dim: events_per_bin} | da.sizes
+    val = sc.broadcast(sc.values(data) / float(events_per_bin), sizes=sizes)
+    kwargs = {'dims': sizes.keys(), 'values': val.values, 'unit': data.unit}
+    if data.variances is not None:
+        kwargs['variances'] = sc.broadcast(
+            sc.variances(data) / float(events_per_bin), sizes=sizes
+        ).values
```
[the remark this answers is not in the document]
nvaytet: I don't know if it really matters in this case? We are just making some fake events that are then meant to be histogrammed. The result of the histogramming should then have the same variances as the original.

I see it as not really an actual step in the reduction (like a normalization) that introduces correlations, but a way to transform the data into a format that can be used further. But maybe I am missing something?

I guess it's important to get this right, because right now this is what would be used for monitor data.

src/ess/reduce/time_of_flight/toa_to_tof.py
nvaytet: No index because the distance is a single number.

src/ess/reduce/time_of_flight/types.py
```diff
@@ -8,37 +8,35 @@
...
+    The results should be flat lists (1d arrays) of the properties of the neutrons in
+    the simulation.
+
+    Parameters
+    ----------
+    time_of_arrival:
+        Time of arrival of the neutrons at the position where the events were recorded.
+        For a ``tof`` simulation, this is just the position of the component (chopper or
+        detector) where the events are recorded. For a ``McStas`` simulation, this is
+        the position of the event monitor.
```
SimonHeybrock, line 31: Was the second part meant for the `distance:` field?

src/ess/reduce/time_of_flight/toa_to_tof.py
nvaytet: I added this to answer a comment from @jokasimr, I meant to investigate later.
I did a preliminary quick benchmark with about 1.5M events (I know it's not so many).
Timings were:

* Running the `tof` simulation: ~2s
* Lookup table computation: 0.30 s
* Scipy Interpolation took 0.19 s

So it's quite fast, but if we do compute the simulation and lookup table only once, it will become the slowest part of the workflow. I need to see how fast it gets with many events.

nvaytet: I tested the scipy `RegularGridInterpolator` a bit more, and it seems it can do about 4-4.5 M float values per second (on my laptop).
It doesn't seem to change if the grid is linspace or not (it's probably doing a `searchsorted` or something internally).

I think leaving the TODO in for now is fine?

jokasimr: I definitely think we should optimize this, but not necessarily in this PR.

1. It's going to be used in every workflow.
2. The cost ~1s per 5M events is quite high. (What's the expected event rate for a typical DREAM experiment? 1e7 counts / second?)
3. It's likely we can make it significantly faster with relatively little effort by (a) reducing the size of the lookup table and by (b) utilizing multithreading.

nvaytet: > reducing the size of the lookup table

I don't think we can make the table smaller really. It's covering a user-specified range in distance (with a user given resolution), and I don't think we should make the toa resolution worse?

So I think we can definitely use multi-threading, and more importantly use the fact that we know the grid is linspace in both dimensions to find indices, instead of having to resort to searchsorted (which I assume is what scipy is using internally).

jokasimr: We can make the lookup table smaller by (for example) fitting piecewise polynomials to it.

That's similar to what we did in the old code, but easier this time because we're not fitting to a polygon but just to a collection of points.

jokasimr: But yes, this is also a good approach:

>  use the fact that we know the grid is linspace in both dimensions to find indices, instead of having to resort to searchsorted (which I assume is what scipy is using internally).

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -0,0 +1,520 @@
...
+    )
+
+    out = da.copy(deep=False)
+    if out.bins is not None:
+        parts = out.bins.constituents
+        out.data = sc.bins(**parts)
+        parts["data"] = tofs
+        out.bins.coords["tof"] = _bins_no_validate(**parts)
+    else:
+        out.coords["tof"] = tofs
```
SimonHeybrock, line 388: Can use `da.bins.assign_coords` and `da.assign_coords` to avoid some back and forth?

src/ess/reduce/time_of_flight/toa_to_tof.py
SimonHeybrock: > I definitely think we should optimize this, but not necessarily in this PR.
> 
>     1. It's going to be used in every workflow.
> 
>     2. The cost ~1s per 5M events is quite high. (What's the expected event rate for a typical DREAM experiment? 1e7 counts / second?)

It will eventually be 5e7 / second.

nvaytet: I implemented the bilinear interpolation manually using numpy (couldn't use scipp arrays in the end because I needed numpy's smart indexing) using the fact that bins are linspace.
It was marginally faster (1.5s for 10M events as opposed to 1.9s with scipy) but nothing amazing.

I will also try with numba but I feel we will probably have to go the C++ way in the end, and use the things already in place in the histogramming code to find the bin indices.

I delay this to later work.

SimonHeybrock: 👍 for delaying.

Can you share the NumPy code though?

src/ess/reduce/time_of_flight/toa_to_tof.py
nvaytet: I changed it to
```Py
    if da.bins is not None:
        parts = da.bins.constituents
        parts["data"] = tofs
        out = da.bins.assign_coords(tof=_bins_no_validate(**parts))
    else:
        out = da.assign_coords(tof=tofs)
```
Seems to work. Is that what you meant?

SimonHeybrock: Yes, `parts` is just a `dict`.

nvaytet: Hmm, I need `scipp=25.01` to use `da.bins.assign_coords` :-(

src/ess/reduce/time_of_flight/toa_to_tof.py
nvaytet: Turns out that `numba` is 30x faster on my workstation... almost 1e9 events per second.
Maybe it's enough speedup?

```Py
from numba import njit, prange

@njit(boundscheck=False, cache=True, fastmath=True, parallel=True)
def interpolate(x, y, z, xp, yp, out):
    npoints = xp.shape[0]
    x0 = x[0]
    y0 = y[0]
    dx = x[1] - x0
    dy = y[1] - y0
    for i in prange(npoints):
        xx = xp[i]
        yy = yp[i]

        ix = int((xx - x0) / dx)
        iy = int((yy - y0) / dy)
        
        y2 = y[iy + 1]
        y1 = y[iy]
        x2 = x[ix + 1]
        x1 = x[ix]
        
        z11 = z[iy, ix]
        z21 = z[iy, ix + 1]
        z12 = z[iy + 1, ix]
        z22 = z[iy + 1, ix + 1]
        
        x2mx1 = x2 - x1
        y2my1 = y2 - y1
        x2mx = x2 - xx
        xmx1 = xx - x1
        
        x2mxox2mx1 = x2mx / x2mx1
        xmx1ox2mx1 = xmx1 / x2mx1
        
        out[i] = (y2 - yy)/y2my1 * (x2mxox2mx1 * z11 + xmx1ox2mx1 * z21) + (yy - y1)/y2my1 * (x2mxox2mx1 * z12 + xmx1ox2mx1 * z22)

N = 10_000_000
result = np.zeros(N)

%%time
interpolate(x, y, da, xp, yp, result)
```
takes 33ms (the second time you run it).

SimonHeybrock: 👍

src/ess/reduce/time_of_flight/toa_to_tof.py
SimonHeybrock: Seems it was released now.

SimonHeybrock (approved):

PR #169, opened 2025-01-23
Change TofWorkflow API to not mirror the Pipeline api

nvaytet:
Maintaining the full `sciline.Pipeline` API on the `TofWorkflow` class is annoying and brittle.
Instead, we allow public access to the underlying `Pipeline`.


SimonHeybrock (approved):
Looks ok, but I wonder if maintaining a wrapper class (including writing all the tests) is worth it. Have you considered using a plain `Pipeline` with a couple of free functions instead?

nvaytet:
> Have you considered using a plain Pipeline with a couple of free functions instead?

Good suggestion. I was also thinking that maintaining the API of this class is going to be a pain.

src/ess/reduce/time_of_flight/toa_to_tof.py
```diff
@@ -459,7 +459,15 @@ def providers() -> tuple[Callable]:
...
-class TofWorkflow:
+def tof_workflow(
+    simulated_neutrons: SimulationResults,
+    ltotal_range: LtotalRange,
+    pulse_stride: PulseStride | None = None,
+    pulse_stride_offset: PulseStrideOffset | None = None,
+    distance_resolution: DistanceResolution | None = None,
+    toa_resolution: TimeOfArrivalResolution | None = None,
+    error_threshold: LookupTableRelativeErrorThreshold | None = None,
+):
```
SimonHeybrock, line 470:
```suggestion
) -> sciline.Pipeline:
```

SimonHeybrock (approved):

PR #171, opened 2025-01-24
Prune workflow in RollingDetectorView

SimonHeybrock:
Working with sciline.Pipeline turns out to be surprisingly slow. Just creating GenericNeXusWorkflow takes close to 200 ms. Pruning the run and monitor types here saves 100 ms in the follow-up code. Since this is called for every detector, it can can add up to about 1 seconds or so in savings for some instruments. This affects startup time of beamlime services.


jl-wynen (commented):
:+1: for the change. But can't you ext6ract the workflow construction from this function and cache the workflow?

SimonHeybrock:
> 👍 for the change. But can't you ext6ract the workflow construction from this function and cache the workflow?

We could, but since it is not a huge problem (done only at startup) I'd rather avoid the complication and instead look into fixing the underlying problem in/via Sciline (see discussion on Slack).

nvaytet (approved):

PR #172, opened 2025-01-29
Fix `to_events` when bins in the input have unsorted edges

nvaytet:


src/ess/reduce/time_of_flight/to_events.py
```diff
@@ -54,6 +54,8 @@ def to_events(
         nans = np.isnan(low) | np.isnan(high)
         low = np.where(nans, 0.0, low)
         high = np.where(nans, 0.0, high)
```
SimonHeybrock, line 56: Had trouble understanding the logic around the change. Should `low` and `high` actually be `left` and `right` (bin edge), until the added lines?

nvaytet: It should probably be left and right everywhere...

src/ess/reduce/time_of_flight/to_events.py
```diff
@@ -45,19 +45,21 @@ def to_events(
...
         # The numpy.random.uniform function below does not support NaNs, so we need to
         # replace them with zeros, and then replace them back after the random numbers
         # have been generated.
-        nans = np.isnan(low) | np.isnan(high)
-        low = np.where(nans, 0.0, low)
-        high = np.where(nans, 0.0, high)
+        nans = np.isnan(left) | np.isnan(right)
+        left = np.where(nans, 0.0, left)
+        right = np.where(nans, 0.0, right)
+        # Ensure low <= high
```
SimonHeybrock, line 57:
```suggestion
        # Ensure left <= right
```

SimonHeybrock (approved):

PR #173, opened 2025-01-31
Add ROIFilter

SimonHeybrock:
This will be used by Beamlime to make time-of-arrival histograms of counts within an ROI. Currently this only supports axis-aligned rectangle selections.

### DIFF
--- a/src/ess/reduce/live/raw.py
+++ b/src/ess/reduce/live/raw.py
@@ -528,9 +528,12 @@ def position_noise_for_cylindrical_pixel(
 
 
 def gaussian_position_noise(sigma: PositionNoiseSigma) -> PositionNoise:
+    sigma = sigma.to(unit='m', copy=False)
     size = _noise_size
     position = sc.empty(sizes={'position': size}, unit='m', dtype=sc.DType.vector3)
-    position.values = np.random.default_rng().normal(0, sigma.value, size=(size, 3))
+    position.values = np.random.default_rng(seed=1234).normal(
+        0, sigma.value, size=(size, 3)
+    )
     return PositionNoise(position)
 
 
--- a/src/ess/reduce/live/raw.py
+++ b/src/ess/reduce/live/raw.py
@@ -175,10 +175,8 @@ class Detector:
         self._start = int(self._flat_detector_number[0].value)
         self._stop = int(self._flat_detector_number[-1].value)
         self._size = int(self._flat_detector_number.size)
-        if not sc.issorted(self._flat_detector_number, dim='event_id'):
-            raise ValueError("Detector numbers must be sorted.")
-        if self._stop - self._start + 1 != self._size:
-            raise ValueError("Detector numbers must be consecutive.")
+        self._sorted = sc.issorted(self._flat_detector_number, dim='event_id')
+        self._consecutive = self._stop - self._start + 1 == self._size
 
     @property
     def detector_number(self) -> sc.Variable:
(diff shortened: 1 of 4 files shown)


src/ess/reduce/live/raw.py
```diff
@@ -343,8 +377,32 @@ def get(self, window: int | None = None) -> sc.DataArray:
...
+        Parameters
+        ----------
+        data:
+            Events grouped by pixel ID, given by binned data.
+        """
+        counts = data.bins.size().to(dtype='int32', copy=False)
+        counts.unit = 'counts'
+        self._add_counts(counts)
+
     def add_counts(self, data: Sequence[int]) -> None:
```
SimonHeybrock, line 393: This will not be used by Beamlime anymore, keeping it for backwards compatibility.

src/ess/reduce/live/raw.py
```diff
@@ -455,9 +528,12 @@ def position_noise_for_cylindrical_pixel(
 
 
 def gaussian_position_noise(sigma: PositionNoiseSigma) -> PositionNoise:
+    sigma = sigma.to(unit='m', copy=False)
     size = _noise_size
     position = sc.empty(sizes={'position': size}, unit='m', dtype=sc.DType.vector3)
-    position.values = np.random.default_rng().normal(0, sigma.value, size=(size, 3))
+    position.values = np.random.default_rng(seed=1234).normal(
```
nvaytet, line 534: Is this to turn the centers of the pixel positions into small gaussian clouds of points?
I was thinking maybe it should be `random.uniform` instead to avoid spill-over from one pixel over to the next?

But maybe using a normal distribution is better as it makes the clouds spherical, and they then look the same from any line-of-sight? Using uniform also suggests that the pixels are square in shape, but this is not really true for tubes...

Is the sigma approximately equal to half the width of the pixel?

src/ess/reduce/live/roi.py
```diff
@@ -0,0 +1,114 @@
...
+        DataGroup with dimension names as keys and tuples of low and high values. This
+        can be used to define a band or a rectangle to selected. When low and high are
+        scipp.Variable, the selection is done using label-based indexing. In this case
+        `indices` must be a DataArray with corresponding coordinates.
+    indices:
+        Variable or DataArray with indices to select from. If binned data the selected
+        indices will be returned concatenated into a dense array.
+    """
+    out_dim = 'index'
+    for dim, (low, high) in intervals.items():
```
nvaytet, line 32: At some point, we may be hooking this up to a graphical interface where a user will draw a rectangle by hand.
In some plotting libraries (e.g. MPL), when you request the bounds of a rectangle, the order of the bounds depends on how the user drew the rectangle.

If they started from the left and dragged towards the right, you get `[left, right]` as the bounds.
They they started from the right, you get `[right, left]` instead.
The same goes fro top/bottom.

We could end up in a situation where some of the tuples contain 2 values but the second one is lower than the first. The slicing below would then select nothing.

I recommend we add something like `low, high = sorted(low, high)` to make sure that does not happen further down the line.

src/ess/reduce/live/raw.py
SimonHeybrock: The gaussian option is just a simple alternative to something that uses the pixel shape, see the cylinder case (used for Loki tubes). We may eventually have something similar for other shapes, I think.

Sigma is a parameter than can be set in a config file in Beamlime.

src/ess/reduce/live/roi.py
SimonHeybrock: Good point, done, please have another look.

nvaytet (approved):

PR #182, opened 2025-02-07
`RollingDetectorView`: Support normalization to number of contributing pixel (fractions)

SimonHeybrock:


src/ess/reduce/live/raw.py
```diff
@@ -125,6 +129,31 @@ def input_indices(self) -> sc.DataArray:
...
+    def intersection_weights(self, threshold: float = 0.25) -> sc.DataArray:
+        """
+        Compute weights as the number of points projected to each bin.
+
+        This is useful for removing projection effects from projected data.
+
+        Parameters
+        ----------
+        threshold:
+            Threshold for identifying bins with a low number of intersections. If the
```
jl-wynen, line 141: Intersections of what?

src/ess/reduce/live/raw.py
```diff
@@ -244,6 +277,13 @@ def __init__(
         self._history: sc.DataArray | None = None
         self._cache: sc.DataArray | None = None
         self.clear_counts()
+        self._norm = (
+            self._projection.intersection_weights(
+                threshold=intersection_weight_threshold
+            )
+            if hasattr(self._projection, 'intersection_weights')
+            else None
```
jl-wynen, line 285: This makes the contract of `projection` complicated. In particular because the normalisation requirements are not communicated anywhere in docs or type hints.

Can the normalisation logic be part of the projection?

In general, why is the projection part of the `live` package? It seems to have nothing to do with live data handling and should really be part of some sort of instrument view package.

SimonHeybrock: > This makes the contract of `projection` complicated. In particular because the normalisation requirements are not communicated anywhere in docs or type hints.
> 
> Can the normalisation logic be part of the projection?

It could be. But not all projections have a sensible normalization (`LogicalView` should not need any). And there would be extra cost to normalizing every time a projection is done, instead of just once when getting the accumulated result. Probably not too bad though?

> In general, why is the projection part of the `live` package? It seems to have nothing to do with live data handling and should really be part of some sort of instrument view package.

Because it was not needed anywhere else so far. Once/if there is another use case it could obviously be moved.

jl-wynen: > And there would be extra cost to normalizing every time a projection is done, instead of just once when getting the accumulated result.

Do you do multiple projections before accumulating? Or how isn't it the same amount of work?

SimonHeybrock: > > And there would be extra cost to normalizing every time a projection is done, instead of just once when getting the accumulated result.
> 
> Do you do multiple projections before accumulating? Or how isn't it the same amount of work?

Not before accumulating itself, but yes, we may be projecting and accumulating 10s or even 100s of chunks before getting the accumulated result.

That being said, after having thought this through some more I realized that this particular normalization may simply be one of multiple options. For example, users might want to normalize to the signal from a uniform scatterer, which is only available in the downstream (Beamlime) code. Therefore I need to try and replace the static mechanism here by something more flexible, which will hopefully also address your concern.

src/ess/reduce/live/raw.py
```diff
@@ -317,19 +289,64 @@ def make_roi_filter(self) -> roi.ROIFilter:
...
+                    f'Invalid {weights.sizes=} for {self.detector_number.sizes=}.'
+                )
+            if (
+                isinstance(weights, sc.DataArray)
+                and (detector_number := weights.coords.get('detector_number'))
+                is not None
+            ):
+                if not sc.identical(detector_number, self.detector_number):
+                    raise sc.CoordError("Mismatching detector numbers in weights.")
+                weights = weights.data
```
jl-wynen, line 327: This is missing a branch for when there is no 'detector_number' coord. In that case, `weights` remains a `DataArray` and it doesn't look like the code below can handle this properly.

src/ess/reduce/live/raw.py
```diff
@@ -317,19 +289,64 @@ def make_roi_filter(self) -> roi.ROIFilter:
...
+        if isinstance(self._projection, Histogrammer):
+            xs = self._projection.apply_full(weights)  # Use all replicas
+        elif self._projection is not None:
+            xs = self._projection(weights)
+        else:
+            xs = weights.copy()
```
jl-wynen, line 333: This now tightly couples `Logicalview` to `Histogrammer`.

I am unclear about what a projection is here. Does it know about replicas in general? Or does only `Histogrammer` know about them? This looks to me like it would lead to inconsistent results between different projections. Can `apply_full` be part of the projection protocol?

SimonHeybrock: > This now tightly couples `Logicalview` to `Histogrammer`.

How so?

> I am unclear about what a projection is here. Does it know about replicas in general? Or does only `Histogrammer` know about them? This looks to me like it would lead to inconsistent results between different projections. Can `apply_full` be part of the projection protocol?

Only the histogrammer. Replicas are not needed in other cases. In what sense inconsistent?

> Can `apply_full` be part of the projection protocol?

It could, but testing is a bit more cumbersome since I cannot pass a simple function as projection (in existing tests unrelated to this change).

src/ess/reduce/live/raw.py
SimonHeybrock: 👍

src/ess/reduce/live/raw.py
jl-wynen: Sorry, not `LogicalView` but `RollingDetectorView`. I unfolded the wrong lines in the diff.

Ok. I am worried that this becomes confusing and complicated as the code grows. But if you can keep it contained and an implementation detail, it should be fine.

jl-wynen (approved):

Issue #184, opened 2025-02-11
Find a reference time for computing pulse_index


nvaytet:
See https://github.com/scipp/essreduce/pull/180#discussion_r1949052455 for context.

When doing pulse-skipping, We need a reliable way to define a reference start time of the run, so we can determine to which pulse in a stride the events are coming from.
The reference time needs to be fixed, in case the data comes in in chunks (cannot simply use the min event_time_offset for each new piece of data).
Some pulses might have no events, so pulses are not always consecutive.

Using a global reference like epoch may work, but it seems difficult to know only from that if the filewriter started recording in the middle of a frame, or at the start of a frame.

It seems the only way would be to look at the open/close times of the pulse-skipping chopper and determine pulse indices from that.
This sounds complicated as we would need to flag which is the pulse-skipping chopper, propagate the open-close times back to the source, and assign pulse indices. The open-close times would need to be in the global timing reference, probably using the chopper TDC time stamps...

SimonHeybrock:
How about the run start time recorded in the NeXus file?

nvaytet:
From discussion with Johannes:

The main thing is that we identify a pulse time when the chopper system has the same exact state as the chopper system in the simulation. That can be done purely from reading the chopper logs, therefore we don't need to trace any neutrons do this.

When determining if the chopper systems states are equal we don't have to specify which chopper is the pulse-skipping chopper, we can just make sure that they all have the right state. If we do it that way we don't have to specify what chopper is "pulse-skipping"

So for example, we could do it this way:

1. Go through all the choppers in the simulation, note how many pulse times from the start time their first tdc time is. For all choppers except the pulse skipping choppers this number will be 0. For the pulse skipping choppers this number is less than the number of skipped pulses.
2. Go through the pulses in the measurement until you find a pulse time where: For each chopper, the number of pulse periods to the next tdc is equal to the corresponding number in the simulation.


So for step 1, you probably wouldn't really need to go through any timings of choppers, you could just do `n_pulses_until_next_tdc = (1.0 / chopper.frequency) // pulse_period`

And for step 2, it means we need to also stream the chopper logs in any live data processing.
In live data, it does seem that the chopper logs get streamed alongside the events, continuously, in a separate topic.

nvaytet:
For now, we have used a workaround in #186 which makes a guess by trying with a small number of events.
This seems to work well, but it is not clear whether it works in all cases.
We may still need to look at chopper data in the long run.

PR #186, opened 2025-02-12
Time-of-flight: use epoch reference time and guess the PulseStrideOffset

nvaytet:
In this PR, we use epoch to create a reference time for computing `pulse_index`.
The method is
```Py
# Define a common reference time using epoch as a base, but making sure that it
# is aligned with the pulse_period and the frame_period.
diff_to_epoch = (event_time_zero.min() - epoch) % pulse_period
# Here we offset the reference by half a pulse period to avoid errors from
# fluctuations in the event_time_zeros in the data. They are triggered by the
# neutron source, and may not always be exactly separated by the pulse period.
# While fluctuations will exist, they will be small, and offsetting the times
# by half a pulse period is a simple enough fix.
reference = epoch + diff_to_epoch - (pulse_period // 2)
# Use in-place operations to avoid large allocations
pulse_index = event_time_zero - reference
pulse_index %= frame_period
pulse_index //= pulse_period
```

We also add a simple method to guess the `PulseStrideOffset` parameter, as using a global reference time now means that the `pulse_index` will depend on when we began to record the data (at the start of a frame period, or half way through it).

We choose a few random events, compute the time-of-flight for every possible value
of `pulse_stride_offset`, and return the value that yields the least number of NaNs
in the computed time-of-flight.

We leave the option to specify the `PulseStrideOffset` as a parameter, and the guess is only performed if it is not provided.

This is a workaround for #184 . It should be easy enough to swap it out later for a solution that reads the chopper logs.


jokasimr:
> In this PR, we use epoch to create a reference time for computing pulse_index

This might be a silly question, but what is `pulse_index` and why do we need it?

Isn't `epoch` just 0?

nvaytet:
> Isn't epoch just 0?

Are you asking why do I need to subtract epoch from the minimum event_time_zero?
I can't do `da.bins.coords['event_time_zero'].min() % pulse_period` because the operation isn't handled event if `pulse_period` is an integer.
Subtracting epoch converts the time stamp to a duration, to which I can then apply the modulo.

jokasimr:
Is it the same as:
```
pulse_index = event_time_zero - min_event_time_zero + pulse_period/2
pulse_index %= frame_period
pulse_index //= pulse_period
```
or am I missing something?

If possible I think it's simpler to not mix in the concept of an `epoch`.
This is more clearly saying: we assume min_event_time_zero coincides with the start of a pulse skipping frame.

nvaytet:
We can't simply use the minimum time because of https://github.com/scipp/essreduce/pull/180#discussion_r1946398402

jokasimr:
> We can't simply use the minimum time because of #180 (comment)

That makes sense 👍 

So essentially the solution is: Instead of assuming the minimum event time zero coincides with a pulse skipping frame start, we assume that the first pulse time after `epoch` coincides with a pulse skipping frame start?

jokasimr:
Aha I see. Pulse index 0 doesn't necessarily meant it's the first pulse of a pulse skipping frame 👍

jokasimr (approved):
Looks good to me!

PR #188, opened 2025-02-12
Refactor unwrap tests

nvaytet:
Change tests to use the choppers from the 'frame unwrapping' docs notebook which offer a wider span in wavelength coverage.

We also refactor to reduce code duplication.

Finally, we also add a test for a pulse stride of 3.


nvaytet:
I know this sort of goes against the advice in the Google book about writing self-contained tests, but I do think it makes it easier to understand what is going on in each test.

SimonHeybrock (approved):

PR #192, opened 2025-02-14
Make live reduction work with BIFROST

MridulS:


src/ess/reduce/live/raw.py
```diff
@@ -639,7 +639,7 @@ def position_with_noisy_replicas(
     # "Paint" the short array of noise on top of the (replicated) position data.
     noise = sc.concat(
         [position_noise] * ceil(size / position_noise.size), dim=noise_dim
-    )[:size].fold(dim=noise_dim, sizes={'replica': replicas, position.dim: -1})
+    )[:size].fold(dim=noise_dim, sizes={'replica': replicas, **position.sizes})
```
SimonHeybrock, line 642: Looks ok, but not that these projections probably do not make much sense for Bifrost, since the postitions of the detector tubes are after scattering from analyzers. How you tried `LogicalView`?

SimonHeybrock (approved):

src/ess/reduce/live/raw.py
MridulS: Hmm, but don't I also need to read in the nexus files to apply the transformations? Or I can just directly use the 2d detectors? I am setting up the config like this right now https://github.com/scipp/beamlime/pull/320

SimonHeybrock: For a first iteration I think we are ok without loading a file. Only once we have files that have calibrated positions that may make sense.

PR #193, opened 2025-02-17
StreamProcessor: Fix dynamic value reuse, allow partial updates

### LINKED ISSUE #191: Likely bug in `StreamProcessor`
`streaming.StreamProcessor.accumulate` loops of the passed chunks, then performs all accumulations. However, if only a subset of dynamic keys is found in the given chunks, previous values appear to be reused, which is likely invalid in most cases. We need to either check the keys, or change the interface in another way.

SimonHeybrock:
Fixes #191. We allow partial updates that are accumulated independently.

### DIFF
--- a/tests/streaming_test.py
+++ b/tests/streaming_test.py
@@ -96,9 +96,11 @@ def test_rolling_accumulator_does_not_modify_pushed_values() -> None:
 
 DynamicA = NewType('DynamicA', float)
 DynamicB = NewType('DynamicB', float)
+DynamicC = NewType('DynamicC', float)
 StaticA = NewType('StaticA', float)
 AccumA = NewType('AccumA', float)
 AccumB = NewType('AccumB', float)
+AccumC = NewType('AccumC', float)
 Target = NewType('Target', float)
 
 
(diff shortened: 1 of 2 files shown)


tests/streaming_test.py
```diff
@@ -322,3 +328,103 @@ def _make_target(accum_a: AccumA, accum_b: AccumB) -> Target:
...
+    streaming_wf = streaming.StreamProcessor(
+        base_workflow=base_workflow,
+        dynamic_keys=(DynamicA, DynamicB),
+        target_keys=(Target,),
+        accumulators=(AccumA, AccumB),
+    )
+    result = streaming_wf.add_chunk({DynamicA: sc.scalar(1), DynamicB: sc.scalar(4)})
+    assert sc.identical(result[Target], sc.scalar(2 * 1.0 / 4.0))
+    result = streaming_wf.add_chunk({DynamicA: sc.scalar(2)})  # Only A
+    assert not sc.identical(result[Target], sc.scalar(2 * 3.0 / 8.0))
```
nvaytet, line 348: Not sure why we need to check that the result is not equal to `sc.scalar(2 * 3.0 / 8.0)` here, if we are checking against the correct expected value just below?

nvaytet (approved):

tests/streaming_test.py
SimonHeybrock: "documenting" what happened before the bug was fixed, otherwise not much.

PR #198, opened 2025-02-26
skip some locking mismatched test on macos and windows

MridulS:
A quick way to avoid https://github.com/scipp/essreduce/issues/197, I've not dug too deep into differences of file locking implementation on MacOS vs Linux, but it looks like on MacOS the implementation is a bit lax?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     needs: formatting
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


jl-wynen:
Is it more lax? Judging by the tests you disabled, it seems like on MacOS locking defaults to `True` because the conflicts are gone. On Linux it defaults to `"best-effort"` which causes the flag conflict that the tests check.

jl-wynen (approved):

PR #201, opened 2025-02-27
Increase diff threshold for histogram unwrap tests

### LINKED ISSUE #200: Flaky unwrapping tests
Some frame unwrapping tests can randomly fail. I observed it locally with these tests:
```
tests/time_of_flight/unwrap_test.py:179 test_standard_unwrap_histogram_mode[tof-80.0]
tests/time_of_flight/unwrap_test.py:179 test_standard_unwrap_histogram_mode[time_of_flight-80.0]
```
The reported problem is
```
>       assert np.nanpercentile(diff.values, percentile) < diff_threshold
E       assert np.float64(0.3026880000000014) < 0.3
```

nvaytet:
Fixes #200


jl-wynen:
I'd like to understand how the results can be different between runs if you seed the rng. Did you check whether the code uses an unseeded rng somewhere? Or maybe the problem is caused by parallelism?

nvaytet:
@jl-wynen have you come across this again recently?
Should we use the updated threshold or should we close this?

I still don't know where this could come from. My guess is that you're right about parallelism, but that's just a guess...

jl-wynen:
I did not encounter this again. I am fine with merging this. The threshold is arbitrary anyway.

jl-wynen (approved):

PR #203, opened 2025-02-28
Add `sum` option to live.raw.LogicalView

SimonHeybrock:
This complements the `slice` capabilities.

### DIFF
--- a/src/ess/reduce/live/raw.py
+++ b/src/ess/reduce/live/raw.py
@@ -157,22 +157,28 @@ class LogicalView:
     select:
         Dimensions with associated index to select from the data. This extracts a slice
         of the data for each given dimension.
+    sum:
+        Dimensions to sum over after folding and selecting, but before flattening.
+        If None, sum over all dimensions.
     flatten:
         Dimensions to flatten.
     """
 
     fold: dict[str, int] | None = None
     transpose: tuple[str, ...] | None = None
-    select: dict[str, int] = field(default_factory=dict)
+    select: dict[str, int | slice] = field(default_factory=dict)
+    sum: str | Sequence[str] | None = ()
     flatten: dict[str, list[str]] = field(default_factory=dict)
 
     def __call__(self, da: sc.DataArray) -> sc.DataArray:
         if self.fold is not None:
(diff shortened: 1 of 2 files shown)


YooSunYoung (approved):
Looks nice...!

Would people also want `mean` instead of `sum` ...?

And maybe it'll be useful to document the order of the operation applied to the data in the docstring as well.

src/ess/reduce/live/raw.py
```diff
@@ -154,25 +154,32 @@ class LogicalView:
         corresponds to multiple dimensions in the logical view.
     transpose:
         Dimensions to transpose. This is useful for reordering dimensions.
+        Applied after sum but before flatten operations.
     select:
         Dimensions with associated index to select from the data. This extracts a slice
         of the data for each given dimension.
+    sum:
+        Dimensions to sum over after folding and selecting, but before transposing and
+        flattening. If None, sum over all dimensions.
```
jl-wynen, line 163: This seems quite complicated, especially because `LogicalView` uses these operations in a fixed order which may be surprising to a user. Did you consider using a user-defined function instead? The operations performed here are all quite simple to express using Scipp. So a function wouldn't be much more verbose. To clarify, I mean something like this:
```python
logical_view = LogicalView(lambda da: da.['a', 0].sum('c').transpose(('d', 'b')))
```

jl-wynen: > Would people also want mean instead of sum ...?

One more reason to let users provide a function.

SimonHeybrock: Works for now as is. Unless you have a good reason for wanting to change it?

YooSunYoung: I think we should keep it just as it is since we will have another service that allows arbitrary workflow to run right...?

jl-wynen: Just keep it in mind for when you add the next operation.

SimonHeybrock: Well, probably `LogicalView` should just be removed, and the `projection` argument of `RollingDetectorView` can take any callable instead.

PR #205, opened 2025-02-28
Add minimum and maximum accumulator.

YooSunYoung:
We needed this accumulator in the `essnmx`, mcstas workflow.

### DIFF
--- a/src/ess/reduce/streaming.py
+++ b/src/ess/reduce/streaming.py
@@ -185,11 +185,10 @@ class MinAccumulator(Accumulator):
         self._cur_min: sc.Variable | None = None
 
     def _do_push(self, value: sc.Variable) -> None:
-        new_min = value.min()
         if self._cur_min is None:
-            self._cur_min = new_min
+            self._cur_min = value
         else:
-            self._cur_min = min(self._cur_min, new_min)
+            self._cur_min = min(self._cur_min, value)
 
     @property
     def is_empty(self) -> bool:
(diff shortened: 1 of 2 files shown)


src/ess/reduce/streaming.py
```diff
@@ -174,6 +174,66 @@ def clear(self) -> None:
         self._values = []
 
 
+class MinAccumulator(Accumulator):
+    """Keeps the minimum value seen so far.
+
+    Only support scalar values.
```
jl-wynen, line 180:
```suggestion
    Only supports scalar values.
```
Same for `MaxAccumulator`.

But this is not actually true. `_do_push` calculates the min / max of the value. So this supports any shape of data.

SimonHeybrock: I thought it would raise for non-scalars?

YooSunYoung: It will raise, because it's just using python built-in `min` and `max`.
Should we filter it in advance...?

jl-wynen: No? The code uses operations along the lines of
```python
min(sc.data.table_xyz(10).data.min(), sc.scalar(10, unit='K'))
```
which works fine.

YooSunYoung: > But this is not actually true. _do_push calculates the min / max of the value. So this supports any shape of data.

No, it raises if you pass 2 non-scalar variables in the `min` and `max`.

jl-wynen: But that is not what your code does. It always passes two results of `.min()` or `.max()` to the builtin `min` and `max` functions. So the inputs will always be scalar.

YooSunYoung: It shouldn't do `.min` and `.max` on the input value since it'll be surprising if the values have different shape with/without the accumulator.

jl-wynen (approved):

src/ess/reduce/streaming.py
SimonHeybrock: Not a request to change, but note that one could use `scipp.reduce([a,b]).min()` to preserve input value shapes.

PR #209, opened 2025-03-03
Add failing test for min/max accumulator.

YooSunYoung:
Related to #205 

I meant to add these tests.

### DIFF
--- a/tests/streaming_test.py
+++ b/tests/streaming_test.py
@@ -164,9 +164,9 @@ def test_min_accumulator() -> None:
         _ = accum.value
 
 
-@pytest.mark.parametrize("accum", [streaming.MinAccumulator, streaming.MaxAccumulator])
-def test_accumulator_non_scalar_raises(accum) -> None:
-    accum = accum()
+@pytest.mark.parametrize("Accum", [streaming.MinAccumulator, streaming.MaxAccumulator])
+def test_accumulator_non_scalar_raises(Accum) -> None:
+    accum = Accum()
     var = sc.array(dims=['x'], values=[1.0, 2.0, 3.0, 2.0, 1.0])
     accum.push(var)  # First push does not raise
     with pytest.raises(sc.DimensionError, match="Expected 0 dimensions"):
--- a/tests/streaming_test.py
+++ b/tests/streaming_test.py
@@ -164,8 +164,9 @@ def test_min_accumulator() -> None:
         _ = accum.value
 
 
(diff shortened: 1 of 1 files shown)


tests/streaming_test.py
```diff
@@ -164,6 +164,14 @@ def test_min_accumulator() -> None:
         _ = accum.value
 
 
+def test_min_accumulator_non_scalar_raises() -> None:
+    accum = streaming.MinAccumulator()
+    var = sc.array(dims=['x'], values=[1.0, 2.0, 3.0, 2.0, 1.0])
+    accum.push(var)  # First push does not raise
+    with pytest.raises(sc.DimensionError, match="Expected 0 dimensions"):
+        accum.push(var)
```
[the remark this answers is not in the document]
YooSunYoung: This test would have prevented my mistake that I accidentally called `min` or `max` function on the pushed variable object.

If you want to add support, you have to write a more complicated accumulator, which is out of scope of this accumulator so it's irrelevant in my opinion.
But even if we want to update these accumulators later, deleting/updating one unit test won't add too much of work.

tests/streaming_test.py
```diff
@@ -164,6 +164,14 @@ def test_min_accumulator() -> None:
         _ = accum.value
 
 
+def test_min_accumulator_non_scalar_raises() -> None:
+    accum = streaming.MinAccumulator()
```
MridulS, line 168:
```suggestion
@pytest.mark.parametrize("accum", [streaming.MinAccumulator, streaming.MaxAccumulator])
def test_min_accumulator_non_scalar_raises(accum) -> None:
    accum = accum()
```

tests/streaming_test.py
```diff
@@ -176,6 +184,14 @@ def test_max_accumulator() -> None:
...
 
 
+def test_max_accumulator_non_scalar_raises() -> None:
+    accum = streaming.MaxAccumulator()
+    var = sc.array(dims=['x'], values=[1.0, 2.0, 3.0, 2.0, 1.0])
+    accum.push(var)  # First push does not raise
+    with pytest.raises(sc.DimensionError, match="Expected 0 dimensions"):
+        accum.push(var)
+
+
```
MridulS, line 194:
```suggestion

```

This can be removed by using the fixture above.

nvaytet:
@YooSunYoung what is the status on this one? Should we get it merged?

nvaytet (approved):

PR #210, opened 2025-03-03
SteamProcessor with "context" changes

### LINKED ISSUE #207: StreamProcessor: Support parameter changes
Currently `StreamProcessor` pre-computes everything it can, i.e., independent of dynamic keys. This is sufficient for basic operation. There are however many cases where sporadically some parameter such as the detector position or the sample angle is updated. We want to avoid treating these as dynamic keys since the computational overhead can be substantial. At the same time, changing these parameters will require re-computing parts of the workflow that are currently fixed after initialization.

SimonHeybrock:
Fixes #207.

Missing a lot of sanity checks and tests.


src/ess/reduce/streaming.py
```diff
@@ -280,16 +299,32 @@ def __init__(
...
-        for key, value in base_workflow.compute(parents).items():
+        nodes = _find_descendants(workflow, dynamic_keys + context_keys)
+        last_static = _find_parents(workflow, nodes) - nodes
+        for key, value in base_workflow.compute(last_static).items():
             workflow[key] = value
 
+        # Nodes that may need updating on context change but should be cached otherwise.
+        dynamic_nodes = _find_descendants(workflow, dynamic_keys)
+        parents_of_dynamic = _find_parents(workflow, dynamic_nodes) - dynamic_nodes
+        context_dependent_parents = parents_of_dynamic & _find_descendants(
```
jl-wynen, line 317: What about target keys that do not depend on dynamic keys? E.g., if the workflow produces an average temperature but that is not used for processing any data. I think that average temperature would not get updated when the temperature context updates.

src/ess/reduce/streaming.py
```diff
@@ -323,6 +357,26 @@ def __init__(
...
+        needs_recompute = set()
+        for key in context:
+            if key not in self._context_keys:
+                raise ValueError(f"Key '{key}' is not a context key")
+            needs_recompute |= self._context_dependencies[key]
+        for key, value in context.items():
+            self._context_workflow[key] = value
+        results = self._context_workflow.compute(needs_recompute)
+        for key, value in results.items():
+            self._process_chunk_workflow[key] = value
```
jl-wynen, line 378: I don't completely follow the graph operations. Does `self._context_dependencies` contain all nodes that depend on the context? Or only those nodes that are direct parents to dynamic or output nodes? If the former, this would cache a lot of data.

tests/streaming_test.py
```diff
@@ -548,3 +548,61 @@ def test_StreamProcessor_clear() -> None:
...
+        base_workflow=wf,
+        dynamic_keys=(Streamed,),
+        context_keys=(Context,),
+        target_keys=(Output,),
+        accumulators=(ProcessedStreamed,),
+    )
+    assert make_static.call_count == 1
+    assert process_context.call_count == 0
+    streaming_wf.set_context({Context: sc.scalar(3)})
+    assert process_context.call_count == 1
```
jl-wynen, line 592: Not sure it is good that setting the context computes intermediates. This means that setting the context is expensive. Can this be done when streamed data is added?

I have two concerns about the current solution:
- Is there a possibility that `set_context` gets called while `accumulate` is running? If so, modifying the workflow while it is computing (more specifically, while it is building the task graph) may lead to inconsistent results. So this approach requires synchronisation between context changes and adding data to accumualte.
- The user of `StreamProcessor` might expect that only accumulating is expensive. So they may write code that a call to `set_context` is cheap. And this can create unexpected lags and difficult-to-understand performance problems.

src/ess/reduce/streaming.py
```diff
@@ -280,16 +299,32 @@ def __init__(
...
+            workflow, context_keys
+        )
+        graph = workflow.underlying_graph
+        self._context_dependencies = {
+            context_key: nx.descendants(graph, context_key) & context_dependent_parents
+            for context_key in self._context_keys
+            if context_key in graph
+        }
+
+        self._context_workflow = workflow.copy()
```
jl-wynen, line 327: This is already an issue with dynamic keys, but is there a way to avoid working on the workflow and instead working with the task graph? This would avoid having to rebuild the task graph on every update.

tests/streaming_test.py
```diff
@@ -548,3 +548,61 @@ def test_StreamProcessor_clear() -> None:
...
+    streaming_wf.accumulate({Streamed: sc.scalar(4)})
+    assert sc.identical(streaming_wf.finalize()[Output], sc.scalar(2 * 3 + 4))
+    assert make_static.call_count == 1
+    assert process_context.call_count == 1
+    streaming_wf.accumulate({Streamed: sc.scalar(4)})
+    assert process_context.call_count == 1
+    streaming_wf.set_context({Context: sc.scalar(1)})
+    # Context changed but no chunk added, so new context is not visible in the result
```
jl-wynen, line 600: What if the result depends only on context but not on a dynamic key?

tests/streaming_test.py
SimonHeybrock: - StreamProcessor is not intended to be usable in a multi-threaded context, so not to the first question.
- I think this is only a concern if the context changes more often that the streamed data, which we probably need to to prevent in practice (filter noise?), or come up with a different solution. Regarding lag, it is actually important to to processing early, as soon as information is available. Say updating context takes 0.5 seconds, and we typically push streamed data twice per second. If we delay the context update until the next data push, you will experience the full 0.5 seconds lag. If you do it immediately some (unpredictable) amount of it can be hidden in the gap between streamed updates.

src/ess/reduce/streaming.py
SimonHeybrock: Hmm, great question. It would require setting values on task graphs. Should be possible in principle, and should be considered if performance benchmarks indicate we should.

src/ess/reduce/streaming.py
SimonHeybrock: Only the direct parents, i.e., it computes as far "down" in the graph as possible.

src/ess/reduce/streaming.py
SimonHeybrock: I think this could actually be useful, I will add support.

src/ess/reduce/streaming.py
SimonHeybrock: I renamed the variable, does that make it clear enough?

tests/streaming_test.py
SimonHeybrock: Added support and a separate test.

tests/streaming_test.py
jl-wynen: This assumes that `T(context_update + stream_update) = T(context_update) + T(stream_update)`, where `T` is run time.
But it is fine, I just wanted to make sure you considered this.

src/ess/reduce/streaming.py
SimonHeybrock: Made a start [here](https://github.com/scipp/essreduce/tree/streaming-no-graph-rebuild), but it was getting slightly too complicated. Waiting for data to decide if it is worth pursuing.

PR #215, opened 2025-03-14
Require a source_position argument when simulating tof beamlines

### LINKED ISSUE #208: Compute chopper positions relative to source
The time of flight workflow requires distances of choppers from the source and computes them as `ch.axle_position.fields.z`. This is only valid if the axle position is relative to the source. That is, the source is the origin of the coordinate system. However, in NeXus files, the origin of the coordinate system is often the sample. So we need to convert compute the distance differently.

I propose to add a `source_position` argument to `simulate_beamline` and compute the distance along the lines of
```python
distance = sc.norm(ch.axle_position - source_position).
```
But note that this is the full euclidean distance. We may need to only consider the z component or in some other way handle the fact that the beam does not hit the axle but the disk.

jl-wynen:
Fixes #208

### DIFF
--- a/src/ess/reduce/time_of_flight/simulation.py
+++ b/src/ess/reduce/time_of_flight/simulation.py
@@ -10,6 +10,8 @@ from .types import SimulationResults
 
 def simulate_beamline(
     choppers: Mapping[str, DiskChopper],
+    source_position: sc.Variable,
+    *,
     neutrons: int = 1_000_000,
     pulses: int = 1,
     seed: int | None = None,
@@ -25,6 +27,9 @@ def simulate_beamline(
         A dict of DiskChopper objects representing the choppers in the beamline. See
         https://scipp.github.io/scippneutron/user-guide/chopper/processing-nexus-choppers.html#Build-DiskChopper
         for more information.
+    source_position:
+        A scalar variable with ``dtype=vector3`` that defines the source position.
+        Must be in the same coordinate system as the choppers' axle positions.
     neutrons:
         Number of neutrons to simulate.
     pulses:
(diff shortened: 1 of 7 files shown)


src/ess/reduce/time_of_flight/simulation.py
```diff
@@ -10,6 +10,8 @@
 
 def simulate_beamline(
     choppers: Mapping[str, DiskChopper],
+    source_position: sc.Variable,
+    *,
```
nvaytet, line 14:
```suggestion
    *,
    choppers: Mapping[str, DiskChopper],
    source_position: sc.Variable,
```

use kwargs only?

tests/time_of_flight/unwrap_test.py
```diff
@@ -178,7 +182,10 @@ def test_unwrap_with_no_choppers() -> None:
         distance=distance,
         choppers=choppers,
         simulation=time_of_flight.simulate_beamline(
-            choppers=choppers, neutrons=300_000, seed=1234
+            choppers=choppers,
+            source_position=sc.vector([0, 0, 0], unit='m'),
```
nvaytet, line 186:
```suggestion
            source_position=fakes.source_position(),
```

src/ess/reduce/time_of_flight/simulation.py
jl-wynen: The first arg has a distinct type, but sure.

tests/time_of_flight/unwrap_test.py
jl-wynen: I didn't do this because `fakes.source_position` corresponds to `fakes.*_choppers` and we don't use those functions here. But I guess the source position actually does not matter here.

PR #216, opened 2025-03-14
Add fast numba regular grid interpolator

nvaytet:
If `numba` is installed on the system, a fast multi-threaded interpolator for time-of-flight is used.
If it is not, it falls back to the current Scipy interpolator.

Fixes scipp/ess#85

### DIFF
--- a/src/ess/reduce/time_of_flight/interpolator_numba.py
+++ b/src/ess/reduce/time_of_flight/interpolator_numba.py
@@ -57,6 +57,7 @@ def interpolate(
     one_over_dx = 1.0 / dx
     one_over_dy = 1.0 / dy
     one_over_dz = 1.0 / dz
+    norm = one_over_dx * one_over_dy * one_over_dz
 
     for i in prange(npoints):
         xx = xp[i]
@@ -94,20 +95,22 @@ def interpolate(
             a122 = values[iz + 1, iy + 1, ix]
             a222 = values[iz + 1, iy + 1, ix + 1]
 
-            x2mxox2mx1 = (x2 - xx) * one_over_dx
-            xmx1ox2mx1 = (xx - x1) * one_over_dx
-            y2myoy2my1 = (y2 - yy) * one_over_dy
-            ymy1oy2my1 = (yy - y1) * one_over_dy
-            z2mzoz2mz1 = (z2 - zz) * one_over_dz
-            zmz1oz2mz1 = (zz - z1) * one_over_dz
-
-            out[i] = z2mzoz2mz1 * (
-                y2myoy2my1 * (x2mxox2mx1 * a111 + xmx1ox2mx1 * a211)
-                + ymy1oy2my1 * (x2mxox2mx1 * a121 + xmx1ox2mx1 * a221)
-            ) + zmz1oz2mz1 * (
-                y2myoy2my1 * (x2mxox2mx1 * a112 + xmx1ox2mx1 * a212)
-                + ymy1oy2my1 * (x2mxox2mx1 * a122 + xmx1ox2mx1 * a222)
-            )
+            x2mxx = x2 - xx
+            xxmx1 = xx - x1
+            y2myy = y2 - yy
+            yymy1 = yy - y1
+            out[i] = (
+                (z2 - zz)
+                * (
+                    y2myy * (x2mxx * a111 + xxmx1 * a211)
+                    + yymy1 * (x2mxx * a121 + xxmx1 * a221)
+                )
+                + (zz - z1)
+                * (
+                    y2myy * (x2mxx * a112 + xxmx1 * a212)
+                    + yymy1 * (x2mxx * a122 + xxmx1 * a222)
+                )
+            ) * norm
 
 
 class Interpolator:


nvaytet:
Any good suggestions for running the tests in a parametrized way so that it tests both implementations?

src/ess/reduce/time_of_flight/interpolator_numba.py
```diff
@@ -0,0 +1,156 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+import numpy as np
+from numba import njit, prange
+
+
+@njit(boundscheck=False, cache=True, fastmath=True, parallel=True)
```
SimonHeybrock, line 7: Is `fastmath` needed? Might be dangerous?

src/ess/reduce/time_of_flight/eto_to_tof.py
```diff
@@ -353,19 +357,26 @@ def __call__(
...
-            if np.any(values < vmin) or np.any(values > vmax):
-                raise ValueError(
-                    "Some requested values are outside of lookup table bounds for "
-                    f"axis {i}: {name}, min: {vmin}, max: {vmax}."
-                )
+            raise ValueError(
+                "Some requested ltotal values are outside of lookup table bounds: "
+                f"min: {self._distance_edges[0]}, max: {self._distance_edges[-1]}."
+            )
+
```
SimonHeybrock, line 374: There is also a bounds check in the interpolator, can that be simplified if it is prevented here anyway?

SimonHeybrock (commented):
How many events/second can this do now?

SimonHeybrock:
> Any good suggestions for running the tests in a parametrized way so that it tests both implementations?

Test the interpolators?

src/ess/reduce/time_of_flight/eto_to_tof.py
```diff
@@ -307,20 +305,26 @@ def __init__(self, lookup: sc.DataArray, distance_unit: str, time_unit: str):
...
-                np.sort(np.concatenate([base_grid - 0.25, base_grid + 0.25])),
-                lookup.coords["distance"].to(unit=distance_unit, copy=False).values,
-                lookup.coords["event_time_offset"]
-                .to(unit=self._time_unit, copy=False)
-                .values,
-            ),
-            np.repeat(
+        self._pulse_edges = np.sort(
+            np.concatenate([base_grid - 0.25, base_grid + 0.25])
+        )
```
jl-wynen, line 310: What is going on here? Where does the value of 0.25 come from? Can't you use the midpoints of the arange from above as edges? Or do you need to have gaps in between the bins?

src/ess/reduce/time_of_flight/interpolator_numba.py
jl-wynen, line 7: Are you sure about `boundscheck=False`? There is no check in this function that the arrays are all the same size. Is this needed for performance?

src/ess/reduce/time_of_flight/interpolator_numba.py
```diff
@@ -0,0 +1,156 @@
...
+            or (yy > ymax)
+            or (zz < zmin)
+            or (zz > zmax)
+        ):
+            out[i] = fill_value
+
+        else:
+            ix = int((xx - xmin) / dx)
+            iy = int((yy - ymin) / dy)
+            iz = int((zz - zmin) / dz)
```
jl-wynen, line 72:
```suggestion
            ix = (xx - xmin) // dx
            iy = (yy - ymin) // dy
            iz = (zz - zmin) // dz
```
Or does numba bot support this? Or does it round in the wrong way?

jl-wynen (commented):
About testing: How about making the interpolator an argument of `TofInterpolator`? That way you can pass whichever interpolator you want. And by default, it can pick one itself.
This would also allow passing more arguments to the interpolator. E.g., the fill value or out of bounds behaviour.

nvaytet:
> How many events/second can this do now?

@jl-wynen reported getting about 29M events / s, but I don't know how many threads this was with.

src/ess/reduce/time_of_flight/eto_to_tof.py
nvaytet: This was added because JL ended in a situation where he was getting all NaNs in his `tof` values, and spending a lot of time trying to debug, only to find out at the end that the range in distance for his lookup table was too narrow and he was sampling outside of the table.

We wanted to warn users about this, so that they could identify the issue faster.
That said, we didn't want to raise on all out of bounds errors because for the `event_time_offset` dimension there could be bins (most likely with 0 counts) that are outside of the 71ms range, in the case of histogrammed data for example. So we still need to check inside the interpolator and return NaNs if we are outside of the table.

I agree it's not very nice to have this check in two places, but not quite the same check. Maybe I should just remove the checks here?
I could add a warning message at the end of the computation that looks at the total number of nans in the output: if there are more than half of nans in the output, warn about needing to check the bounds of the table?

src/ess/reduce/time_of_flight/eto_to_tof.py
nvaytet: There is no need for gaps. I tried to explain it in the comment just above these lines. This has been in there for a while. The number 0.25 is just arbitrary. I could have chosen `[-0.49, 0.49, 0.51, 1.49]`. I just cannot have `[-0.5, 0.5, 0.5, 1.5]` (the double 0.5 in the middle). The first 2 values need to be on either side of 0, and the next 2 values on either side of 1, and so on.

src/ess/reduce/time_of_flight/interpolator_numba.py
nvaytet: It was for perfomance, but we can put it back if you like. I can also disable `fastmath` as @SimonHeybrock requested above.

src/ess/reduce/time_of_flight/interpolator_numba.py
nvaytet: It should be fine with your suggestion. They are equivalent, aren't they?

src/ess/reduce/time_of_flight/eto_to_tof.py
SimonHeybrock: All this will impact performance somehow, so it should be further analyzed (not necessarily here).

jl-wynen: > I agree it's not very nice to have this check in two places, but not quite the same check. Maybe I should just remove the checks here?

If you remove the check here and make the interpolator raise, then it will also raise for out of bounds times. And from what I remember, you wanted to not raise in this case in order to support monitors.

src/ess/reduce/time_of_flight/interpolator_numba.py
jl-wynen: Not for negative numbers:
```python
>>> int(-3/2)
-1
>>> -3//2
-2
```

src/ess/reduce/time_of_flight/eto_to_tof.py
nvaytet: I just removed those extra checks for now, and if we get all NaNs, we just need to know where to look.
Checking all the values was potentially quite expensive.

src/ess/reduce/time_of_flight/interpolator_numba.py
nvaytet: Ah yes, but I don't think it matters in this case as we are always taking `(x - xmin)` which should always be positive or 0. If it is negative, it was caught by the `if xx < xmin` further up.

SimonHeybrock:
> > > How many events/second can this do now?
> > 
> > 
> > @jl-wynen reported getting about 29M events / s, but I don't know how many threads this was with. He basically got a speedup of 10x
> 
> Hmmm, might be ok for now, but we likely need to improve that more, somehow. Sounds like even with the speedup this may be a bottleneck for the entire data reduction?

Was curios and tested *just* the Numba function on my machine. Not sure what params to use but, e.g., for `nx ,ny, nz = 2000, 150, 10` it processes 100M events in 250 ms, which is 10x higher that what @jl-wynen got (presumably for the entire mechanism). It should be worth investigating whether the bottleneck lies elsewhere now.

src/ess/reduce/time_of_flight/interpolator_numba.py
```diff
@@ -0,0 +1,159 @@
...
+            a121 = values[iz, iy + 1, ix]
+            a221 = values[iz, iy + 1, ix + 1]
+            a112 = values[iz + 1, iy, ix]
+            a212 = values[iz + 1, iy, ix + 1]
+            a122 = values[iz + 1, iy + 1, ix]
+            a222 = values[iz + 1, iy + 1, ix + 1]
+
+            x2mx1 = x2 - x1
+            y2my1 = y2 - y1
+            z2mz1 = z2 - z1
```
SimonHeybrock, line 95: If `x`, `y`, and `z` are linspace, why do we compute the deltas over and over again? Seems like a waste since they should all be the same? Can then also compute inverse just once, so we can scale, which is faster.

src/ess/reduce/time_of_flight/interpolator_numba.py
```diff
@@ -0,0 +1,159 @@
...
+            z2mzoz2mz1 = (z2 - zz) / z2mz1
+            zmz1oz2mz1 = (zz - z1) / z2mz1
+
+            out[i] = z2mzoz2mz1 * (
+                y2myoy2my1 * (x2mxox2mx1 * a111 + xmx1ox2mx1 * a211)
+                + ymy1oy2my1 * (x2mxox2mx1 * a121 + xmx1ox2mx1 * a221)
+            ) + zmz1oz2mz1 * (
+                y2myoy2my1 * (x2mxox2mx1 * a112 + xmx1ox2mx1 * a212)
+                + ymy1oy2my1 * (x2mxox2mx1 * a122 + xmx1ox2mx1 * a222)
+            )
```
SimonHeybrock, line 110: Can avoid the 6 divisions and long variables names sth. like
```python
x2mx = (x2 - xx)
...
norm = z2mz1 * y2my1 * x2mx1  # outside loop?
...
out[i] = (existing code) / norm
```
`norm` could be absorbed into `values` I suppose, so we can completely avoid handling this? The edges and values never change independently, do they?

nvaytet: Ah because `z2mzoz2mz1 * (y2myoy2my1 * (x2mxox2mx1 ...` would result in `z2mz1 * y2my1 * x2mx1` on the denominator :-)
:+1:

nvaytet:
I will perform full benchmarks of the time-of-flight computation in separate work.
I think we do not need to wait for that to get these changes merged, as they already provide significant speedup.

SimonHeybrock (approved):

SimonHeybrock:
Where should Numba be added to dependencies? In all the technique packages? While optional, I think not having it by default will lead to a lot of surprisingly bad performance for users.

PR #223, opened 2025-04-02
Avoid integer overflow in time-of-flight calculation

nvaytet:
Convert `event_time_offset` internally to floats to avoid integer overflows when computing tof.

### DIFF
--- a/src/ess/reduce/time_of_flight/eto_to_tof.py
+++ b/src/ess/reduce/time_of_flight/eto_to_tof.py
@@ -371,7 +371,8 @@ def _time_of_flight_data_histogram(
     # In NeXus, 'time_of_flight' is the canonical name in NXmonitor, but in some files,
     # it may be called 'tof'.
     key = next(iter(set(da.coords.keys()) & {"time_of_flight", "tof"}))
-    eto_unit = da.coords[key].unit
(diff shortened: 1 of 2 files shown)


src/ess/reduce/time_of_flight/eto_to_tof.py
```diff
@@ -371,17 +371,16 @@ def _time_of_flight_data_histogram(
     # In NeXus, 'time_of_flight' is the canonical name in NXmonitor, but in some files,
     # it may be called 'tof'.
     key = next(iter(set(da.coords.keys()) & {"time_of_flight", "tof"}))
-    eto_unit = da.coords[key].unit
+    raw_eto = da.coords[key].to(dtype=float, copy=False)
```
jl-wynen, line 374: Can this cause rounding errors when we have very large numbers but care about small intervals? E.g., because we have nanoseconds since epoch but frames lengths on the order of milliseconds?

nvaytet: Actually don't have nanoseconds from epoch, it's only `event_time_offset`, to max is 71ms.
I realise now it's not failing because of overflow, as it also fails with int64.
It's failing for all ints, because the `numba` interpolator expects floats.

jl-wynen (approved):

PR #227, opened 2025-04-09
Use Scope from sciline as ScopeTwoParams is deprecated.

MridulS:


jl-wynen:
This will only work when we release Sciline. I am actually going to change it to not deprecate `ScopeTwoParams` yet. So that we can properly migrate and don't have a lot of broken builds.

jl-wynen:
I just released sciline. So let's revive this PR.
I also released essreduce so that we don't force the update of sciline on users right away.

jl-wynen (approved):

PR #228, opened 2025-04-09
Fix time-of-flight numba interpolation

nvaytet:
There was an out-of-bounds issue when a value is exactly on the right edge of the table.
This value is introduced manually when data is histogrammed so that we can rebin the data exactly where the table wraps around.

### DIFF
--- a/src/ess/reduce/time_of_flight/interpolator_numba.py
+++ b/src/ess/reduce/time_of_flight/interpolator_numba.py
@@ -43,13 +43,16 @@ def interpolate(
     if not (len(xp) == len(yp) == len(zp) == len(out)):
         raise ValueError("Interpolator: all input arrays must have the same size.")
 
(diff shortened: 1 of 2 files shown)


jokasimr (commented):
(slight tangent) Am I understanding it correctly that we're doing a 3D interpolation where one of the dimensions is `pulse`? In that case it seems quite odd to me. Time of flight is not continuous in pulse so it doesn't make sense to "interpolate between pulses". I guess the results are still correct, because the pulse variable always exactly equals one if the pulse edges? But that means
  * we're actually doing a 2D interpolation, but 'paying' for a 3D interpolation in terms of compute
  * if the value is 'nan' in the next pulse that is going to propagate to the previous pulse, even though they ought to be unrelated.

Regarding the actual change in the PR: LGTM 👍

PR #230, opened 2025-04-10
Time-of-flight: remove pulse dimension in lookup table

### LINKED ISSUE #229: Time-of-flight: Change 3D interpolation to 2D by making a table that wraps at frame_period
(slight tangent) Am I understanding it correctly that we're doing a 3D interpolation where one of the dimensions is `pulse`? In that case it seems odd to me. I guess the results are still correct, because the pulse variable always exactly equals one if the pulse edges? But that means
  * we're actually doing a 2D interpolation, but 'paying' for a 3D interpolation in terms of compute
  * if the value is 'nan' in the next pulse that is going to propagate to the previous pulse, even though they ought to be unrelated. 

Regarding the actual change in the PR: LGTM 👍

_Originally posted by @jokasimr in https://github.com/scipp/essreduce/pull/228#pullrequestreview-2754008587_

nvaytet:
Instead of making a 3D lookup table with a `pulse` in the case of pulse skipping, we just make a table which spans the `frame_period` instead of the `pulse_period`.
We now send `event_time_offset + pulse_period * pulse_index` as the x coordinates to the interpolator.

This has simplified quite a number of things in the implementation.

Fixes #229 

TODO: I still need to re-run the benchmarks to see if this increased performance (hopefully it did).

### DIFF
--- a/src/ess/reduce/time_of_flight/eto_to_tof.py
+++ b/src/ess/reduce/time_of_flight/eto_to_tof.py
@@ -166,6 +166,43 @@ def compute_tof_lookup_table(
     error_threshold:
         Threshold for the relative standard deviation (coefficient of variation) of the
         projected time-of-flight above which values are masked.
+
+    Notes
+    -----
+
+    Below are some details about the binning and wrapping around frame period in the
+    time dimension.
+
+    We have some simulated ``toa`` (events) from a Tof/McStas simulation.
+    Those are absolute ``toa``, unwrapped.
+    First we compute the usual ``event_time_offset = toa % frame_period``.
+
+    Now, we want to ensure periodic boundaries. If we make a bin centered around 0,
+    and a bin centered around 71ms: the first bin will use events between 0 and
+    ``0.5 * dt`` (where ``dt`` is the bin width).
+    The last bin will use events between ``frame_period - 0.5*dt`` and
+    ``frame_period + 0.5 * dt``. So when we compute the mean inside those two bins,
+    they will not yield the same results.
+    It is as if the first bin is missing the events it should have between
+    ``-0.5 * dt`` and 0 (because of the modulo we computed above).
+
+    To fix this, we do not make a last bin around 71ms (the bins stop at
+    ``frame_period - 0.5*dt``). Instead, we compute modulo a second time,
+    but this time using ``event_time_offset %= (frame_period - 0.5*dt)``.
+    (we cannot directly do ``event_time_offset = toa % (frame_period - 0.5*dt)`` in a
+    single step because it would introduce a gradual shift,
+    as the pulse number increases).
+
+    This second modulo effectively takes all the events that would have gone in the
+    last bin (between ``frame_period - 0.5*dt`` and ``frame_period``) and puts them in
+    the first bin. Instead of placing them between ``-0.5*dt`` and 0,
+    it places them between 0 and ``0.5*dt``, but this does not really matter,
+    because we then take the mean inside the first bin.
+    Whether the events are on the left or right side of zero does not matter.
+
+    Finally, we make a copy of the left edge, and append it to the right of the table,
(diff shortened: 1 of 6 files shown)


src/ess/reduce/time_of_flight/eto_to_tof.py
```diff
@@ -276,39 +230,31 @@ def compute_tof_lookup_table(
...
-        table=table, pulse_period=pulse_period, pulse_stride=pulse_stride
+    # Copy the left edge to the right to create periodic boundary conditions
+    table = sc.DataArray(
+        data=sc.concat(
+            [table.data, table.data['event_time_offset', 0]], dim='event_time_offset'
+        ),
+        coords={
+            "distance": table.coords["distance"],
+            "event_time_offset": sc.concat(
+                [table.coords["event_time_offset"], frame_period],
```
jokasimr, line 241: Is `sc.concat([table.coords["event_time_offset"], frame_period])` guaranteed to be equidistant?

Maybe it's better to just not include the endpoint. I mean, on the lhs we don't start with 0, why do we have to end with `frame_period` on the rhs?

jokasimr (commented):
Looks good! I only have one question, besides that I'm all for merging.

src/ess/reduce/time_of_flight/eto_to_tof.py
nvaytet: > Is sc.concat([table.coords["event_time_offset"], frame_period]) guaranteed to be equidistant?

I think so, from the construction... at least to machine precision.

> Maybe it's better to just not include the endpoint. I mean, on the lhs we don't start with 0, why do we have to end with frame_period on the rhs?

We cannot do without the end point, if not you lose events at the edge.
On the left edges we do begin at exactly 0. Why do you think we don't?

jokasimr: > On the left edges we do begin at exactly 0. Why do you think we don't?

Because we are using `sc.midpoints(table.coords["event_time_offset"])` and I assumed `table.coords["event_time_offset"][0] == 0` before we call `midpoints`.

> I think so, from the construction... at least to machine precision.

Aha I see, the original `event_time_offset` grid is from `-bin_width/2` to `frame_period - bin_width/2`.

How about instead we just create the original grid from `-bin_width/2` to `frame_period + bin_width/2`?
Then when we do `midpoints` on that we get the grid we want without gluing the first value to the end.

nvaytet: > How about instead we create the original grid from -bin_width/2 to frame_period + bin_width/2?

If you do this, you get a weird effect around 71ms.
Screenshot_20250410_121831

It's complicated to explain without a whiteboard, but the events that are just above 71ms, in the second half of the top bin, should not be included in the mean for that bin. Instead we wrap them around to the left edge, and then copy the left edge to the right, to ensure perfectly periodic boundaries.

This is the reason we have [this bit of code](https://github.com/scipp/essreduce/blob/main/src/ess/reduce/time_of_flight/eto_to_tof.py#L111-L118).

jokasimr: Aha, the tof simulation will still return events in that region. I see.

In `_compute_mean_tof_in_distance_range`, can we not do `data.coords['event_time_offset'] %= frame_period` instead and use the grid from `-bin_width/2` to `frame_period + bin_width/2`.

That seems simpler to me. But maybe it doesn't work?

jokasimr: Thanks for the explanation, it's clear now. :+1:

jokasimr (approved):

src/ess/reduce/time_of_flight/eto_to_tof.py
nvaytet: I should probably add this explanation in the docstring or something

PR #231, opened 2025-04-14
Time-of-flight interpolator optmizations

nvaytet:
Avoid some large allocations when not needed.

The 2D implementation was actually slower than the old one because of allocations.
We now got the time for ~72M events down from 0.83s to 0.33s.
The old 3D implementation is at 0.43s.

~Unfortunately, I had to duplicate code in the numba interpolator.~
~Suggestions for a better alternative welcome.~
Fixed it, you can just use the `None` type in Numba, I wasn't sure you could...

### DIFF
--- a/src/ess/reduce/time_of_flight/eto_to_tof.py
+++ b/src/ess/reduce/time_of_flight/eto_to_tof.py
@@ -311,7 +311,8 @@ class TofInterpolator:
         self,
         ltotal: sc.Variable,
         event_time_offset: sc.Variable,
-        pulse_offset: sc.Variable | None = None,
+        pulse_period: sc.Variable,
+        pulse_index: sc.Variable | None = None,
     ) -> sc.Variable:
         if ltotal.unit != self._distance_unit:
             raise sc.UnitError(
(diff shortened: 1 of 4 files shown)


jokasimr:
Did you figure out exactly what allocations made the performance difference?

We had two allocations here:
```
event_time_offset=etos + pulse_index * pulse_period.to(unit=eto_unit),
```

but in this PR we also introduce one allocation here
```
pulse_offset = pulse_index * pulse_period.to(unit=eto_unit)
``` 
so as far as I can see the difference between the old and the new version is just one allocation, is that right?
It seems unlikely we would see such a difference because of a single extra allocation.

But if that's the case, couldn't we achieve the same result by eliminate the extra allocation like this?
```
_eto = pulse_index * pulse_period.to(unit=eto_unit, dtype=etos.dtype)
_eto += etos
...
# in the call to `interp`
event_time_offset=_eto,
```

jokasimr:
Another option is to pass `pulse_index` and `pulse_period` directly to the interpolation function (instead of `pulse_offset`). Then we don't need the `pulse_offset` array.

nvaytet:
> Did avoiding to construct the pulse_offset array make any difference performance-wise?

In my test case, with ~72M events, run-time went from 0.87s to 0.77s, so it's non-negligible.

jokasimr:
What's the difference between that benchmark and the one you posted earlier? There we got 0.33s for 72M events.

nvaytet:
> What's the difference between that benchmark and the one you posted earlier? There we got 0.33s for 72M events.

The difference is between pulse skipping and not.
It apparently takes 0.36s to compute the pulse index, without even doing the guesswork (to find the `PulseStrideOffset`).
Would be good if we could bring that down...

nvaytet:
I'll merge for now, as at least it's faster than the previously merged implementation.

PR #233, opened 2025-04-28
Generic tof workflow

nvaytet:
We build on top of the `GenericNexusWorkflow` to add the computation of time-of-flight for the neutron data.

See https://github.com/scipp/essimaging/pull/81 for an example of how this is used in instrument reduction workflows.

The graph is the following. Important thing to note is that Ltotal is extracted from the calibrated beamline, not from the detector data, meaning that it is not extracted/computed again every time new streaming events come in.
tbl-tof-wf

### DIFF
--- a/src/ess/reduce/nexus/types.py
+++ b/src/ess/reduce/nexus/types.py
@@ -232,7 +232,7 @@ class CalibratedDetector(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
     """Calibrated data from a detector."""
 
 
-class CalibratedBeamline(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
+class CalibratedDetectorBeamline(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
     """Calibrated beamline with detector and other components."""
 
 
@@ -240,6 +240,12 @@ class CalibratedMonitor(
     sciline.Scope[RunType, MonitorType, sc.DataArray], sc.DataArray
 ):
     """Calibrated data from a monitor."""
+
+
+class CalibratedMonitorBeamline(
+    sciline.Scope[RunType, MonitorType, sc.DataArray], sc.DataArray
+):
+    """Calibrated beamline with monitor and other components."""
 
 
 class DetectorData(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
--- a/src/ess/reduce/nexus/workflow.py
+++ b/src/ess/reduce/nexus/workflow.py
@@ -20,9 +20,10 @@ from .types import (
     AllNeXusComponents,
     Analyzers,
     Beamline,
-    CalibratedBeamline,
     CalibratedDetector,
+    CalibratedDetectorBeamline,
     CalibratedMonitor,
+    CalibratedMonitorBeamline,
     Choppers,
     Component,
     DetectorBankSizes,
@@ -401,12 +402,12 @@ def get_calibrated_detector(
     )
 
 
-def assemble_beamline(
+def assemble_detector_beamline(
     detector: CalibratedDetector[RunType],
     source_position: Position[snx.NXsource, RunType],
     sample_position: Position[snx.NXsample, RunType],
     gravity: GravityVector,
-) -> CalibratedBeamline[RunType]:
+) -> CalibratedDetectorBeamline[RunType]:
     """
     Add beamline information (gravity vector, source- and sample-position) to detector.
 
@@ -427,7 +428,7 @@ def assemble_beamline(
     gravity:
         Gravity vector.
     """
-    return CalibratedBeamline[RunType](
+    return CalibratedDetectorBeamline[RunType](
         detector.assign_coords(
             source_position=source_position,
             sample_position=sample_position,
@@ -437,7 +438,7 @@ def assemble_beamline(
 
 
 def assemble_detector_data(
-    detector: CalibratedBeamline[RunType],
+    detector: CalibratedDetectorBeamline[RunType],
     event_data: NeXusData[snx.NXdetector, RunType],
 ) -> DetectorData[RunType]:
     """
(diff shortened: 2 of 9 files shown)


src/ess/reduce/time_of_flight/workflow.py
```diff
@@ -0,0 +1,30 @@
...
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+from collections.abc import Iterable
+
+import sciline
+
+from ..nexus import GenericNeXusWorkflow
+from .eto_to_tof import default_parameters, providers
+
+
+def GenericTofWorkflow(
```
nvaytet, line 11: Reviewer: should we update the docs to use this notebook, or should we leave theme as low-level, manually constructing the pipeline?

See for example [here](https://scipp.github.io/essreduce/user-guide/tof/dream.html#Setting-up-the-workflow):
Screenshot_20250428_141055

src/ess/reduce/nexus/types.py
```diff
@@ -232,7 +232,7 @@ class CalibratedDetector(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
     """Calibrated data from a detector."""
 
 
-class CalibratedBeamline(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
+class CalibratedDetectorBeamline(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
```
SimonHeybrock, line 235: We spoke the other day about not breaking downstream and notebooks anymore by renaming (unless really necessary). Is this necessary? Is the new name better?

src/ess/reduce/nexus/workflow.py
```diff
@@ -491,6 +492,34 @@ def get_calibrated_monitor(
...
+    gravity: GravityVector,
+) -> CalibratedMonitorBeamline[RunType, MonitorType]:
+    """
+    Add beamline information (gravity vector, source- and sample-position) to monitor.
+
+    This is performed separately and after :py:func:`get_calibrated_monitor` to avoid
+    as false dependency of, e.g., the reshaped detector numbers on the sample position.
+    The latter can change during a run, e.g., for a rotating sample. The detector
+    numbers might be used, e.g., to mask certain detector pixels, and should not depend
+    on the sample position.
```
SimonHeybrock, line 507: This does not apply here? Is there thus any reason to not perform this in `get_calibrated_monitor`? Would avoid the extra type and the "need" to rename `CalibratedBeamline`.

src/ess/reduce/nexus/workflow.py
```diff
@@ -718,33 +748,8 @@ def GenericNeXusWorkflow(
     wf[PreopenNeXusFile] = PreopenNeXusFile(False)
 
     if run_types is not None or monitor_types is not None:
-        _prune_type_vars(wf, run_types=run_types, monitor_types=monitor_types)
-
-    return wf
+        from ..utils import prune_type_vars
```
SimonHeybrock, line 751: Why the import here?

src/ess/reduce/time_of_flight/workflow.py
```diff
@@ -0,0 +1,30 @@
...
+    *,
+    run_types: Iterable[sciline.typing.Key] | None = None,
+    monitor_types: Iterable[sciline.typing.Key] | None = None,
+) -> sciline.Pipeline:
+    """ """
+    wf = GenericNeXusWorkflow()
```
SimonHeybrock, line 17: I think this may be slow unless you forward the `run_types` etc.? Not sure if you then need to prune again, or if there is a better way.

SimonHeybrock:
Wondering about this bit: 
image

Why do we need the `All*` types? Can loading the choppers be incorporated into the `NeXusComponentLocationSpec` and `NeXusComponent`? Or can the `All*` types be replaced by a more chopper-specific name, as it seems that is the only thing used?

SimonHeybrock:
Looking at 
image
it seems that branching / feeding `PulsePeriod` and `PulseStride` into both `TimeOfFlightLookupTable` and `DetectorTofData` is error prone. Is there a risk that the wrong table is used? Can the branching be avoided?

nvaytet:
> Is there a risk that the wrong table is used? Can the branching be avoided?

Are you thinking of a case where for example you loaded the lookup table from a file, and then set `PulseStride=2` even if the lookup table was created for a `PulseStride=1`?

SimonHeybrock:
> > Is there a risk that the wrong table is used? Can the branching be avoided?
> 
> Are you thinking of a case where for example you loaded the lookup table from a file, and then set `PulseStride=2` even if the lookup table was created for a `PulseStride=1`?

Yes, something like that.

src/ess/reduce/time_of_flight/workflow.py
nvaytet: I opted for pruning twice.

nvaytet:
> Why do we need the All* types? Can loading the choppers be incorporated into the NeXusComponentLocationSpec and NeXusComponent? Or can the All* types be replaced by a more chopper-specific name, as it seems that is the only thing used?

I am not sure, would be best to discuss this in person/over a call.

nvaytet:
> > > Is there a risk that the wrong table is used? Can the branching be avoided?
> > 
> > 
> > Are you thinking of a case where for example you loaded the lookup table from a file, and then set `PulseStride=2` even if the lookup table was created for a `PulseStride=1`?
> 
> Yes, something like that.

So maybe we should store the `PulsePeriod` and the `PulseStride` inside the table (as extra coordinates, or turn it into a `dataclass`)?

src/ess/reduce/nexus/workflow.py
```diff
@@ -401,7 +402,7 @@ def get_calibrated_detector(
     )
 
 
-def assemble_beamline(
+def assemble_detector_beamline(
```
SimonHeybrock, line 405: Can also revert this rename.

src/ess/reduce/time_of_flight/eto_to_tof.py
```diff
@@ -520,14 +535,57 @@ def _time_of_flight_data_events(
...
-def time_of_flight_data(
-    da: RawData,
+def extract_detector_ltotal(
+    detector_beamline: CalibratedBeamline[RunType],
+) -> DetectorLtotal[RunType]:
+    """
+    Extract Ltotal from the detector data.
+    """
+    graph = scn.conversion.graph.beamline.beamline(scatter=True)
+    return DetectorLtotal[RunType](
```
SimonHeybrock, line 545: I suppose what this really does is the naive straight-line approximation to Ltotal based on basic component positions. I guess the idea is to swap out this provider if something more sophisticated is needed? Maybe the nature of the approximation should be reflected in the names and docstrings?

SimonHeybrock:
> > Why do we need the All* types? Can loading the choppers be incorporated into the NeXusComponentLocationSpec and NeXusComponent? Or can the All* types be replaced by a more chopper-specific name, as it seems that is the only thing used?
> 
> I am not sure, would be best to discuss this in person/over a call.

Conclusion from call: Turns out this was unrelated to this change.

PR #237, opened 2025-05-09
Remove analyzers from nexus workflow

jl-wynen:
Analyzers are fairly uncommon at ESS (3 instruments) and as of right now, we don't load them using the mechanism that was removed here. Instead, we have a dedicated implementation for BIFROST. So this should not break anything downstream.

### DIFF
--- a/src/ess/reduce/nexus/types.py
+++ b/src/ess/reduce/nexus/types.py
@@ -341,10 +341,3 @@ class Choppers(
     sc.DataGroup[sc.DataGroup[Any]],
 ):
     """All choppers in a NeXus file."""
-
-
-class Analyzers(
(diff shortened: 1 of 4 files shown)


SimonHeybrock:
Are you sure you want to remove them, instead of simply not including it in `GenericNeXusWorkflow` by default?

jl-wynen:
We currently have no use for them. (For BIFROST, we load each detector-analyser pair separately.) Loading all analysers in one go like this would require also loading detectors differently in BIFROST. And we don't know how analysers will be stored for MIRACLES and VESPA. So I am fine with removing it and potentially re-adding it if we find that it is useful after all.

SimonHeybrock (approved):

PR #239, opened 2025-05-12
Time-of-flight: Add loading lookup table from file

nvaytet:
`GenericTofWorkflow` will try to load from a file by default. Keyword arg allows for selecting a Tof or McStas simulation (not implemented, but by using an `Enum` adding this should hopefully be possible without breaking changes).


nvaytet (commented):
Looks good, thanks for picking this up.
(Can't approve my own PR 😉)

PR #242, opened 2025-05-14
Distinguish between raw and parsed choppers

jl-wynen:
The tof workflow incorrectly assumes that loaded `Choppers` are `DiskChopper` objects. In reality, they are `DataGroup`s. This means that the tof workflow fails at during `compute`. This PR splits the `Choppers` type into two. One for the nested data groups loaded from NeXus and one for processed `DiskChopper` objects. There is no provider to connect the two because that is a complicated process that involves finding plateaus and extracting scalar quantities from the NeXus data. See https://scipp.github.io/scippneutron/user-guide/chopper/processing-nexus-choppers.html

If the data is completely flat, this can be simple. E.g., for the simulated BIFROST data, we can simply do this:
```python
def extract_chopper_plateau(chopper):
    processed = chopper.copy()
    # These are constant in the simulated data.
    processed['rotation_speed'] = processed['rotation_speed'].data.mean()
    processed['phase'] = processed['phase'].data.mean()
    # Guessing here as this is not stored in the file.
    processed['beam_position'] = sc.scalar(0.0, unit='deg')
    return DiskChopper.from_nexus(processed)


def extract_chopper_plateaus(choppers: RawChoppers[RunType]) -> DiskChoppers[RunType]:
    return DiskChoppers[RunType](choppers.apply(extract_chopper_plateau))

workflow.insert(parse_disk_choppers)
```

### DIFF
--- a/src/ess/reduce/nexus/workflow.py
+++ b/src/ess/reduce/nexus/workflow.py
@@ -23,7 +23,6 @@ from .types import (
     CalibratedBeamline,
     CalibratedDetector,
     CalibratedMonitor,
-    Choppers,
     Component,
     DetectorBankSizes,
     DetectorData,
@@ -46,6 +45,7 @@ from .types import (
     NeXusTransformationChain,
     Position,
     PreopenNeXusFile,
+    RawChoppers,
     RunType,
     SampleRun,
     TimeInterval,
@@ -513,9 +513,18 @@ def assemble_monitor_data(
 
 def parse_disk_choppers(
     choppers: AllNeXusComponents[snx.NXdisk_chopper, RunType],
-) -> Choppers[RunType]:
-    """Convert the NeXus representation of a chopper to ours."""
-    return Choppers[RunType](
+) -> RawChoppers[RunType]:
+    """Convert the NeXus representation of a chopper to ours.
+
+    Returns
+    -------
+    :
+        A nested data group containing the loaded choppers.
+        The elements may be time-dependent arrays that first need to be processed
+        before they can be passed to other functions as
+        :class:`ess.reduce.nexus.types.DiskChoppers`.
+    """
+    return RawChoppers[RunType](
         choppers.apply(
             lambda chopper: extract_chopper_from_nexus(
                 nexus.compute_component_position(chopper)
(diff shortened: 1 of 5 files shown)


SimonHeybrock:
> The tof workflow incorrectly assumes that loaded `Choppers` are `DiskChopper` objects.

So how did the workflow work so far? Have chopper params always been set by hand (using the "incorrect" type)?

src/ess/reduce/nexus/workflow.py
```diff
@@ -513,9 +513,18 @@ def assemble_monitor_data(
 
 def parse_disk_choppers(
     choppers: AllNeXusComponents[snx.NXdisk_chopper, RunType],
-) -> Choppers[RunType]:
-    """Convert the NeXus representation of a chopper to ours."""
-    return Choppers[RunType](
+) -> RawChoppers[RunType]:
```
nvaytet, line 516: > > > The tof workflow incorrectly assumes that loaded Choppers are DiskChopper objects.
> >
> > So how did the workflow work so far? Have chopper params always been set by hand (using the "incorrect" type)?
>
> I think so. I don't see any other way that would have worked. In particular for DREAM, we never (to my knowledge) used nexus files, only a hard coded list of choppers.

I think this is correct, we read and used choppers from nexus for the TBL but in that file, the list of choppers was empty, so we then just built a LUT without choppers.

nvaytet (approved):

jl-wynen:
> > If the data is completely flat, this can be simple. E.g., for the simulated BIFROST data, we can simply do this:
> 
> Note that ECDC has started putting the 'setpoint' values for the rotation speed and phase in the nexus files, which is the target value that the control software is supposed to achieve. So in principle, if we trust that, we could just read that instead of finding plateaus (which would remain useful for debugging). Screenshot_20250422_181511

Good to know! If these setpoints are good enough, we can have a default provider that maps `RawChoppers` to `DiskChoppers`.

SimonHeybrock:
> > > If the data is completely flat, this can be simple. E.g., for the simulated BIFROST data, we can simply do this:
> > 
> > 
> > Note that ECDC has started putting the 'setpoint' values for the rotation speed and phase in the nexus files, which is the target value that the control software is supposed to achieve. So in principle, if we trust that, we could just read that instead of finding plateaus (which would remain useful for debugging). Screenshot_20250422_181511
> 
> Good to know! If these setpoints are good enough, we can have a default provider that maps `RawChoppers` to `DiskChoppers`.

We might also want to avoid loading the complete timestamp logs, since they might get quite big?

PR #244, opened 2025-05-20
Histogram monitors

jl-wynen:
ECDC write monitors as histograms but our workflow only supports event mode monitors. This PR fixes that. 

This should also fix
- https://git.esss.dk/dmsc-nightly/dmsc-nightly/-/issues/30
- https://git.esss.dk/dmsc-nightly/dmsc-nightly/-/issues/31

### DIFF
--- a/src/ess/reduce/nexus/workflow.py
+++ b/src/ess/reduce/nexus/workflow.py
@@ -591,15 +591,19 @@ def _add_variances(da: sc.DataArray) -> sc.DataArray:
     out = da.copy(deep=False)
     if out.bins is not None:
         content = out.bins.constituents['data']
-        if content.variances is None:
-            content.variances = content.values
+        content.data = _assign_values_as_variances(content.data)
     elif out.variances is None:
-        try:
-            out.variances = out.values
-        except sc.VariancesError:
-            out = out.to(dtype=sc.DType.float64)
-            out.variances = out.values
+        out.data = _assign_values_as_variances(out.data)
     return out
+
+
+def _assign_values_as_variances(var: sc.Variable) -> sc.Variable:
+    try:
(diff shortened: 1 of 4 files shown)


src/ess/reduce/nexus/workflow.py
```diff
@@ -590,6 +593,12 @@ def _add_variances(da: sc.DataArray) -> sc.DataArray:
         content = out.bins.constituents['data']
         if content.variances is None:
             content.variances = content.values
```
nvaytet, line 595: Do we also need the dtype check here?

tests/nexus/workflow_test.py
```diff
@@ -519,6 +558,21 @@ def test_load_monitor_workflow() -> None:
...
+
+def test_load_histogram_monitor_workflow() -> None:
+    wf = LoadMonitorWorkflow(run_types=[SampleRun], monitor_types=[FrameMonitor1])
+    wf[Filename[SampleRun]] = data.dream_coda_test_file()
+    wf[NeXusName[FrameMonitor1]] = 'monitor_bunker'
+    da = wf.compute(MonitorData[SampleRun, FrameMonitor1])
+    assert 'position' in da.coords
+    assert 'source_position' in da.coords
+    assert da.bins is None
+    assert da.dims == ('time', 'frame_time')
```
nvaytet, line 572:
```suggestion
    assert set(da.dims) == {'time', 'frame_time'}
```
Could the order of dims change?

nvaytet (commented):
Looks good

tests/nexus/workflow_test.py
jl-wynen: Possibly. And we don't need to rely on a specific order :+1:

PR #246, opened 2025-05-26
Remove separate resampling step from TofWorkflow

SimonHeybrock:
After #245 the resampling step can (I hope) be considered less experimental. This change simplifies the workflow, making downstream code more straightforward by avoiding having to modify the workflows depending on whether event data or histogram data is passed.

Unfortunately this is also a breaking change, requiring refactoring downstream, where the (monitor) resampling was used. Example is scipp/essdiffraction#171.


nvaytet (approved):
🥳

PR #251, opened 2025-07-04
Use `RunType` instead of `SampleRun` for tof lookup table

nvaytet:
See discussion https://github.com/scipp/essimaging/pull/91#discussion_r2180018466 for context.

This should allow more flexible use of run types created in downstream packages, as opposed to only being able to use aliases of the types defined in `essreduce`.

Downside is that this is now a breaking change (again).

There were also other uses of `SampleRun` in the `live` submodule here, but I did not modify those, as I am unsure if it is required.

### DIFF
--- a/docs/user-guide/tof/dream.ipynb
+++ b/docs/user-guide/tof/dream.ipynb
@@ -337,7 +337,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "wf[time_of_flight.SimulationResults] = time_of_flight.simulate_beamline(\n",
+    "wf[time_of_flight.SimulationResults[SampleRun]] = time_of_flight.simulate_beamline(\n",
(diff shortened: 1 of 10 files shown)


jl-wynen:
One option would be to have types like `GenericSimulationResults` and default providers that convert those into `SimulationResults[RunType]`, etc. But this would double the size of the tof graph.

Loading from file is probably fast enough because the tables are fairly small. In practice, that is probably what we would use most of the time, right? We would usually not run a simulation during reduction? So we could just leave it as is and see if we get into performance problems.

nvaytet:
Is there a way to have a function take in `DiskChoppers[RunType]`, inside loop over all the run types and check that all chopper settings are the same, and then return a single `Choppers` (or a better name) which does not depend on `RunType`?

PR #253, opened 2025-07-08
Use a dedicated workflow to compute tof LUT

nvaytet:
This is an alternative to #251 .

Instead of having one tof LUT per `RunType`, we have a single LUT, which is loaded from file.
This will be the main mode of operation for production workflows.

If one wishes to compute the LUT on-the-fly, using `tof`, we now have to use a separate workflow that is designed to do just this. The `TofLutWorkflow` is used to compute `TimeOfFlightLookupTable`, and the result of this is then set on the `GenericTofWorkflow`, bypassing the provider that loads the table from file.

I don't think that having different LUT for different runs is a good idea in practice (@jokasimr pointed out that different chopper settings would change the distribution of neutrons, which would make different measurements incompatible in most cases). 
If it turns out we do need this in the future, we could always add a wrapper provider that turns `TimeOfFlightLookupTable` into `TimeOfFlightLookupTable[RunType]`.
Inserting a custom LUT for a single run can then still be done via
```Py
table = lut_wf.compute(TimeOfFlightLookupTable)
wf[TimeOfFlightLookupTable[SampleRun]] = table
```


src/ess/reduce/time_of_flight/lut.py
```diff
@@ -0,0 +1,481 @@
...
+    return SimulationResults(
+        time_of_arrival=events.coords["toa"],
+        speed=events.coords["speed"],
+        wavelength=events.coords["wavelength"],
+        weight=events.data,
+        distance=furthest_chopper.distance,
+    )
+
+
+def TofLutWorkflow():
```
nvaytet, line 462: Any suggestions for a better name a welcome :smile:

src/ess/reduce/time_of_flight/lut.py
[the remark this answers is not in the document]
nvaytet: Only moved code.

src/ess/reduce/time_of_flight/lut.py
nvaytet: I'm also thinking of renaming `TimeOfFlightLookupTable` to `TofLookupTable`, as part of the naming standardization and freeze (e.g. https://github.com/scipp/ess-project/issues/38).

PR #254, opened 2025-07-11
Drop `event_time_offset` coord after conversion to tof

### LINKED ISSUE #249: Drop `event_time_offset` after computing `tof` in `GenericTofWorkflow`?
Currently this seems to be kept, leading to hidden/surprising size and compute overhead in downstream workflows, unless it is stripped manually.

Can we drop it automatically? Do we need a flag to keep it, if needed for something?

nvaytet:
We also drop the equivalent coords for non-event data.

Note that we keep the `event_time_zero` coord because it can be used for event filtering, based on sample environment or other.

Fixes #249


jl-wynen:
I rebased this on #253 so I can review it.

SimonHeybrock:
What is your plan for computing "time at sample" for high-precision event filtering?

nvaytet:
> Will the TOF workflow be able to compute it on request, before this is dropped?

It would be simpler to compute it by default. But would this be too expensive? (CPU and/or memory)

PR #278, opened 2025-11-11
Add load_field and load_group to allow loading custom entries using GenericNeXusWorkflow

nvaytet:
Example (from the tests):

```Py
class UserAffiliation(sl.Scope[RunType, str], str):
    """User affiliation."""

def load_user_affiliation(
    file: NeXusFileSpec[RunType],
    path: NeXusName[UserAffiliation[RunType]],
) -> UserAffiliation[RunType]:
    return UserAffiliation[RunType](load_field(file, path))

wf = GenericNeXusWorkflow(run_types=[SampleRun], monitor_types=[])
wf.insert(load_user_affiliation)
wf[Filename[SampleRun]] = loki_tutorial_sample_run_60250
wf[NeXusName[UserAffiliation[SampleRun]]] = '/entry/user_0/affiliation'
affiliation = wf.compute(UserAffiliation[SampleRun])
assert affiliation == 'ESS'
```

### DIFF
--- a/tests/nexus/workflow_test.py
+++ b/tests/nexus/workflow_test.py
@@ -776,9 +776,9 @@ def test_generic_nexus_workflow_includes_only_given_monitor_types() -> None:
 
 
 def assert_not_contains_type_arg(node: object, excluded: set[type]) -> None:
-    assert not any(arg in excluded for arg in getattr(node, "__args__", ())), (
-        f"Node {node} contains one of {excluded!r}"
-    )
+    assert not any(
+        arg in excluded for arg in getattr(node, "__args__", ())
+    ), f"Node {node} contains one of {excluded!r}"
 
 
 def test_generic_nexus_workflow_load_custom_field_user_affiliation(


src/ess/reduce/nexus/workflow.py
```diff
@@ -683,23 +683,21 @@ def LoadMonitorWorkflow(
...
         ),
     )
     wf[PreopenNeXusFile] = PreopenNeXusFile(False)
     return wf
 
 
 def LoadDetectorWorkflow(
-    *,
-    run_types: Iterable[sciline.typing.Key],
-    monitor_types: Iterable[sciline.typing.Key],
```
nvaytet: Not sure why we needed to have the monitor_types here, when we are loading detectors?
So I removed it.

tests/nexus/workflow_test.py
```diff
@@ -776,3 +777,53 @@ def assert_not_contains_type_arg(node: object, excluded: set[type]) -> None:
...
+    # Load a component in one of the top-level entries
+
+    class UserAffiliation(sl.Scope[RunType, str], str):
+        """User affiliation."""
+
+    def load_user_affiliation(
+        location: NeXusComponentLocationSpec[UserAffiliation, RunType],
+    ) -> UserAffiliation[RunType]:
+        return UserAffiliation[RunType](
+            load_component(location, nx_class=snx.NXuser)['affiliation']
```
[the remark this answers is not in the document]
SimonHeybrock: `load_component` was made for any "physical component" (it handles positions, etc.). Say maybe it is reasonable and by design that it fails? Which brings us back to the question of what you gain by trying to uses the existing machinery instead of having a simple load provider?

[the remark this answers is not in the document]
nvaytet: I really don't mind either way, I could not figure out how to properly just add a simple load provider which would hook into the part of the workflow that opens the file for reading.

I thought that using the `Filename[SampleRun]` as input to the provider would mean we open the file multiple times, and I wanted to avoid that.
So I would welcome some pointers as to what the provider would take in as input so it can load a custom field. Thanks in advance :slightly_smiling_face:

tests/nexus/workflow_test.py
```diff
@@ -776,3 +783,44 @@ def assert_not_contains_type_arg(node: object, excluded: set[type]) -> None:
...
+    ) -> UserInfo[RunType]:
+        return UserInfo[RunType](load_group(file, path))
+
+    wf = GenericNeXusWorkflow(run_types=[SampleRun], monitor_types=[])
+    wf.insert(load_user_info)
+    wf[Filename[SampleRun]] = loki_tutorial_sample_run_60250
+    wf[NeXusName[UserInfo]] = '/entry/user_0'
+    user_info = wf.compute(UserInfo[SampleRun])
+    assert user_info['affiliation'] == 'ESS'
+    assert user_info['name'] == 'John Doe'
```
jl-wynen: This should ideally be handled similarly to https://github.com/scipp/essreduce/blob/b46edacc7092a9d3babf5a3550625a98587cab35/src/ess/reduce/nexus/_nexus_loader.py#L104
and load one or more https://github.com/scipp/scippneutron/blob/3c49525dd89af68d375119c6f9072008f337dc6c/src/scippneutron/metadata/_model.py#L168

Does your code provide the basis for doing that?

src/ess/reduce/nexus/_nexus_loader.py
```diff
@@ -42,6 +43,58 @@ def __repr__(self) -> str:
...
+        Selection to apply to the field.
+
+    Returns
+    -------
+    :
+        The loaded field as a variable or data array.
+    """
+    with open_nexus_file(filename.value) as f:
+        field = f[field_path]
+        return cast(sc.Variable | sc.DataArray, field[selection])
```
jl-wynen: Does scippnexus guarantee to return a variable even if the dataset is a string?

tests/nexus/workflow_test.py
nvaytet: So are you suggesting that instead of having
```Py
    class UserAffiliation(sl.Scope[RunType, str], str):
        """User affiliation."""

    def load_user_affiliation(
        file: NeXusFileSpec[RunType],
        path: NeXusName[UserAffiliation[RunType]],
    ) -> UserAffiliation[RunType]:
        return UserAffiliation[RunType](load_field(file, path))
```
I would need to create a pydantic model for the `UserAffiliation` with a `from_nexus_entry` classmethod, and then my `load_user_affiliation` would call `load_metadata` instead of `load_field`?

I guess I could but it feels a bit overkill?
I agree that for the proton charge, it might be worth it, because it will be used by multiple instruments.
But for something like an `ExposureTime` which is very specific to imaging, I feel it's too much effort?

In addition, I already felt it was annoying to have to make changes in `essreduce` so I could use them in `essimaging`, now I would have to first change scippneutron, then essreduce, then imaging...
I will say it again: should `scippneutron` and `essreduce` be merged?

jl-wynen: You don't need to make a model for `UserAffiliation`. We already have a `Person` model. All you need is a generic domain type for it (Like your `UserAffiliation`) and a provider to load it. It doesn't have to use a class method in ScippNeutron. And the code that needs the affiliation can extract that from a `Person`.

I think that the generic workflow should be able to load all users and provide a way to select one. (by index, path, or name, probably) Especially because this will be relevant to all workflows in some way.

And please don't misunderstand me, I think it is good to have tools for loading anything from the file in a simple way. But people related info is used in a lot of places. So I think we should have a common, robust, and flexible solution.

nvaytet: So, did I understand correctly that your objection was not adding `load_field` and `load_group`, but it was because I was using them to load the affiliation instead of using the `load_metadata` and `Person` model?

I just picked that field randomly, to illustrate that we can load something custom from the file. Is it better if I pick something else to load?

src/ess/reduce/nexus/_nexus_loader.py
```diff
@@ -42,6 +43,58 @@ def __repr__(self) -> str:
 NoLockingIfNeeded = NoLockingIfNeededType()
 
 
+def load_field(
+    filename: NeXusFileSpec,
+    field_path: str,
+    selection: snx.typing.ScippIndex | slice = (),
```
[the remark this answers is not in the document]
jl-wynen: Can't you just use `NeXusLocationSpec`? It underlies `NeXusComponentLocationSpec` among others. So in a workflow, we only need a single new domain type instead of 2 custom range types as in your example.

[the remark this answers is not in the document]
nvaytet: So if `load_field` takes in a `NeXusLocationSpec` as input, I either have to repeat the filename, e.g.
```Py
wf[Filename[SampleRun]] = loki_tutorial_sample_run_60250()
wf[NeXusLocationSpec[UserAffiliation[SampleRun]]] = NeXusLocationSpec(
    filename=loki_tutorial_sample_run_60250(),
    entry_name='/entry/user_0/affiliation',
    selection=...
)
```
or add a function like `component_spec_by_name` that makes the transition from `NeXusFileSpec` and `NeXusName` to a `NeXusLocationSpec`.

As mentioned above, I would then have to make a new generic `NeXusEntryLocationSpec` which uses a new `TypeVar` not limited to what is covered by `Component`, etc.

> So you may as well use the same mechanism as the existing code.

That's what I tried to explain as to why I didn't use the same mechanism.
But I also said I don't mind implementing that if you think it's better.

jl-wynen (changes requested):

src/ess/reduce/nexus/_nexus_loader.py
```diff
@@ -44,55 +43,47 @@ def __repr__(self) -> str:
...
     -------
     :
-        The loaded field as a variable or data array.
+        The loaded field (as a variable, data array, or raw python object).
     """
-    with open_nexus_file(filename.value) as f:
-        field = f[field_path]
-        return field[selection]
+    with open_nexus_file(location.filename, definitions=definitions) as f:
+        field = f[location.entry_name]
```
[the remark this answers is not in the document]
jl-wynen: > Sorry I still don't understand. In my example above, how would you then load the user affiliation? You can't load a field directly with `_unique_child_group`, it will raise `Expected a NeXus group as item '{name}' but got a field.`

```python
def load_user_affiliation(
    file: NeXusFileSpec[RunType],
    path: NeXusName[UserAffiliation[RunType]],
) -> UserAffiliation[RunType]:
    return UserAffiliation[RunType](load_field(
        NeXusLocationSpec(
                filename=file,
                component_name=path,
                selection=...
        ))

# ...

wf[NeXusName[UserAffiliation[SampleRun]]] = 'user_0/affiliation'
```
where `NeXusName[UserAffiliation[SampleRun]]` is relative to the entry.

jl-wynen (approved):

PR #279, opened 2025-11-13
Save tof lookup table as a dataclass that also contains chopper info

nvaytet:
We change the format of the tof lookup table from a DataArray to a dataclass, which enables us to also save the chopper info as a nested DataGroup (we convert to/from a DataGroup for serialization/deserialization).

The chopper info is crucial to have alongside the table, so we can add validation when trying to use the table, that it applies for the correct chopper settings.

~:warning:  **Note that unfortunately, this will break all workflows that currently use a tof lookup table stored in a hdf file.**~

**Edit:** found a way to make this compatible with the old format, so we don't break existing workflow and we can continue to use existing tables.

See scipp/ess#87 for a discussion as to why in the end I went with scipp.DataGroup/hdf5 instead of a NeXus file.

Needs https://github.com/scipp/scippneutron/pull/649 & https://github.com/scipp/tof/pull/105

### DIFF
--- a/src/ess/reduce/time_of_flight/lut.py
+++ b/src/ess/reduce/time_of_flight/lut.py
@@ -372,23 +372,20 @@ def make_tof_lookup_table(
     # In-place masking for better performance
     _mask_large_uncertainty(table, error_threshold)
 
-    out = {
-        "array": table,
-        "pulse_period": pulse_period,
-        "pulse_stride": pulse_stride,
-        "distance_resolution": table.coords["distance"][1]
-        - table.coords["distance"][0],
-        "time_resolution": table.coords["event_time_offset"][1]
+    return TimeOfFlightLookupTable(
+        array=table,
+        pulse_period=pulse_period,
+        pulse_stride=pulse_stride,
+        distance_resolution=table.coords["distance"][1] - table.coords["distance"][0],
+        time_resolution=table.coords["event_time_offset"][1]
         - table.coords["event_time_offset"][0],
-        "error_threshold": error_threshold,
-    }
-
-    if simulation.choppers is not None:
-        out['choppers'] = sc.DataGroup(
+        error_threshold=error_threshold,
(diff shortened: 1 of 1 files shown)


docs/user-guide/tof/dream.ipynb
```diff
@@ -414,7 +414,7 @@
     "table = lut_wf.compute(TimeOfFlightLookupTable)\n",
     "\n",
     "# Overlay mean on the figure above\n",
-    "table[\"distance\", 13].plot(ax=fig2.ax, color=\"C1\", ls=\"-\", marker=None)"
+    "table[\"data\"][\"distance\", 13].plot(ax=fig2.ax, color=\"C1\", ls=\"-\", marker=None)"
```
SimonHeybrock: Before reviewing further, I really wonder if TimeOfFlightLookupTable should be a (data)class instead of a `DataGroup`? What do you think?

SimonHeybrock: Or maybe even pydantic could be useful here, since it is about ser/deser.

nvaytet: DataGroup makes it very easy to save to hdf5.
What do we gain from using a dataclass instead?

nvaytet: > You can still convert to a DataGroup for ser/deser.

So you would do something like `sc.DataGroup(asdict(table)).save_hdf5('file.h5')` ?
Ok, I'll have a go at that.
I feel pydantic would be a bit overkill.

nvaytet: Any good suggestions for a name for the field I named `"data"`?

nvaytet: I'm going with 'array', I think it's quite common to use that for the thing that actually holds the data.

SimonHeybrock (approved):
👍 (see optional comment)

PR #280, opened 2025-11-20
Fix dream wfm tests

nvaytet:
With the latest release of `tof`, the nightly tests started failing.
It turns out that when computing the wavelengths of neutrons, there was a chance of less than 1 in 10,000 to have some outliers where the error is greater than 2%.

Here is a figure of the error of each neutron in the test, using 100K neutrons. We see 5 outliers.
Figure 1 (64)

I checked and even with the previous version of `tof`, if we use more neutrons than the 10K used in the test, we also start to see the outliers. So it is just a statistical fluctuation and we were simply lucky not to see it before.

To fix, we relax the percentile threshold by a very small amount (100 -> 99.9) to make the test more robust to future changes.

### DIFF
--- a/tests/time_of_flight/wfm_test.py
+++ b/tests/time_of_flight/wfm_test.py
@@ -206,7 +206,7 @@ def test_dream_wfm(
             (x.coords["wavelength"] - ref.coords["wavelength"])
             / ref.coords["wavelength"]
         )
-        assert np.nanpercentile(diff.values, 100) < 0.02
+        assert np.nanpercentile(diff.values, 99.9) < 0.02
         assert sc.isclose(ref.data.sum(), da.data.sum(), rtol=sc.scalar(1.0e-3))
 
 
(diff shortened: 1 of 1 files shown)


nvaytet:
> That strongly indicates there's some change in the last tof release that made this much more likely than before.

I don't think so.
It didn't happen before because we were using 10K neutrons in the tests and we were always using the same seed.
If you use the old version of tof and either change the seed or increase the number of neutrons, outliers can appear.

Something changed in tof for sure: how the pulse is generated.

Old: to select neutrons, do a 1d `random.choice` in a distribution of birth times, and another `random.choice` in a distribution of wavelengths.

New: we have a 2d probability distribution, that depends on birth time and wavelength. We flatten the distribution, do a single `random.choice` on the flattened array, this gives us an index from which we find a birth time and a wavelength.
The 2d distribution was created by broadcasting the old 1d distributions together.

PR #295, opened 2025-12-12
Flat detector viewer

nvaytet:
Add a generic version of the Dream [FlatVoxelViewer](https://github.com/scipp/essdiffraction/blob/main/src/ess/dream/diagnostics.py#L19) so it can be used by other instruments.

The Dream voxel viewer can then be replaced by:
```Py
from ess.reduce.tools.flat_detector_viewer import FlatDetectorViewer

FlatDetectorViewer(dg, horizontal_dim="module", vertical_dim="segment")
```

### DIFF
--- a/src/ess/reduce/tools/__init__.py
+++ b/src/ess/reduce/tools/__init__.py
@@ -1,0 +1,6 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+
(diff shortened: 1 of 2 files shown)


nvaytet:
In the end, this was replaced by the [LokiBankViewer](https://scipp.github.io/esssans/user-guide/loki/loki-detector-diagnostics.html#The-LokiBankViewer) in `esssans` and the changes here are no longer needed.
Please re-open if this can be useful for other instruments.

